PR #12, opened 2015-07-31
Android support + minor improvements

anestisb:
* Add Android API 21 support (make android) using POSIX / SIGNAL arch
* Add support for targets with args of type "arg=___FILE___"
* Add .gitignore

Have tested a few Android ARM targets fuzzing with Lollipop (API 21) and haven't spotted any issues. ptrace and perf counters (ARM & x86) is also doable in Android, although that requires a lot more tweaking and testing due to incompatibilities with bionic ptrace APIs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,10 @@
+*.o
+*.so
+honggfuzz
+*.dSYM
+mach_exc.h
+mach_excUser.c
+mach_excServer.h
+mach_excServer.c
+libs
+obj
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ ifeq ($(OS),Linux)
 		# Support for popcnt (used in linux/perf.c)
 		CFLAGS += -msse4.2
 	endif	# MARCH
-endif	# OS
+endif	# OS Linux
 
 ifeq ($(OS),Darwin)
 	OS_VERSION = $(shell sw_vers -productVersion)
@@ -96,7 +96,7 @@ endif
 	MIG_OUTPUT = mach_exc.h mach_excUser.c mach_excServer.h mach_excServer.c
 	MIG_OBJECTS = mach_excUser.o mach_excServer.o
 	ARCH = DARWIN
-endif
+endif	# OS Darwin
 
 SRCS += $(ARCH_SRCS)
 CFLAGS += -D_HF_ARCH_${ARCH}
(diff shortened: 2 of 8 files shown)


robertswiecki:
Thank you very much for the patch. I'll obviously merge it.

Just, one comment about Android - I think the ptrace interface is so useful (and perf to a lesser degree) that it would, in my opinion, be more beneficial to compile honggfuzz statically on your local Linux machine, and just upload it to the Android machine.

Given that there are support structures for ARM/ARM64 in linux/ptrace.c, it should simply work (unless the Android kernel is so modified, that there are syscall incompatibilities).

anestisb:
I'll give it a try (it's the bfd that would need some engineering as it's not present in Android - will probably require to switch to libbacktrace).

When I find some more time the next bit that I'll clean-up and commit is the mangle callbacks. Roughly the idea is that someone can create a smart mangle template to apply mangling with some custom logic (using exported rng utils_ funcs) and/or possible post mangling actions (e.g. repair CRCs). I've found it super useful for quickly coding some file formats smart fuzzing where file structs (at least the core ones) definitions are available in c/c++ style. If you have any preferences for the gluing logic feel free to email me.

Btw I've used these callbacks to quickly code some dirty PoCs before starting writing another fuzzing engine for Android [1]. So thanks for maintaining this awesome tool.

[1] https://conference.hitb.org/hitbsecconf2015ams/materials/D1T2%20-%20Anestis%20Bechtsoudis%20-%20Fuzzing%20Objects%20d%E2%80%99%20ART.pdf

robertswiecki:
Thanks for the preso link!

FYI: As for the static compilation, I was rather thinking of compiling it statically under a regular Linux box with glibc. Either under an ARM box where it shoild work out of the box, or using cross-compilation, where it might be a bit harder because of libraries you'd have to provide (unwind, libbfd) to the cross-compiler.

When testing ptrace under ARM I compiled honggfuzz under emulator (qemu-system-arm) and it worked nicely. I was wondering if a statically compiled binary will work under Android. IMO it should, given that the Android's kernel is not modified too much from a vanilla Linux ARM kernel.

anestisb:
Well using a generic gnueabi toolchain to static (cross-)compile (including glibc, because of bionic) for Android is never easy since there usually exist a couple of link issues from the deps that need manual resolving. Plus IMHO it's very impractical to debug, specially for a ptrace APIs where you need to carefully debug because of the Android debuggerd ptrace calls that will interfere yours for the worker processes.

It seems that porting the ptrace arch was easier than originally expected. I had to put back a static build of libcapstone due to lack of standalone libbfd that be easily cross-compiled (unless I've missed something out there). Build scripts are configured to only include the matching CPU arch significantly reducing the size of the library.

Have a look at this branch:
https://github.com/anestisb/honggfuzz/tree/master_dev

You can find some notes here:
https://gist.github.com/anestisb/afb85fd90276acc786ab

Everything can be build with official Android NDK toolchains (so far tried gcc only, although clang should be also ok), allowing people to easy debug and build on top. I've tested everything using latest Android NDK r10e for both Linux & OSX (build scripts are configured to detect host OS).

I'll put some notes (specially for the build flags) together and do some more testing before submitting the pull request. Feel free to cherry pick before that.

PR #13, opened 2015-08-20
ptrace() API support for Android

anestisb:
Everything is cross-compiled with official Android NDK toolchains. As such users can build from both Linux & OSX boxes (haven't tested Windows). Has been tested under a few Nexus (4, 5, 6, 10) Lollipop devices and emulator setups.  

libunwind was a bit tricky in some places (and still is for x86_64 ABI), although all core functionality should be covered now. An alternative to manual patches, would be to fetch AOSP libunwind (which is supported by Android dev team), although that would require a lot of Makefile work to build as standalone and not as part of an AOSP build env.

x86_64 ABI is in the todo list, although I need to first resolve some AOSP qemu build issues and setup an emulator env before start investigating the libunwind issues.

Capstone is built with matching arch support. As such the overhead is minimal.

In case we want support for APIs < 21, a workaround for getdelim() usage should be coded as it's not present in Android libc (bionic) for older APIs. However, I don't think it's worth the effort. 

I've also put some notes on the top of the doc/Android.md together here:
https://gist.github.com/anestisb/afb85fd90276acc786ab

Let me know in case you need additional comments or testing for certain bits.


anestisb:
btw I thought that Android CLA (signed long time ago) was covering all project contributions, but apparently it's not the case. So I've signed the other one too to get this covered.

robertswiecki:
Merged! Thanks for taking time and making this happen.

PR #17, opened 2015-09-06
ptrace ARM fix, hardlinks copy failover, dynamic workspace & report file improvements

anestisb:
* GETREGS fix for old ARM kernels
    - PTRACE_GETREGS for some old ARM kernels is failing to extract register values if provided struct size is bigger than expected pt_regs. Bug is fixed by avoiding 32/64-bit structs multiplexing for ARM kernels. arch_getPC() will always execute the 32bit branch so no further pre-processor pollution is required.
    - Bug has been spotted when fuzzing with Nexus4 5.1.1 running 3.4 kernel

* Bypass hardlink restrictions
    - Running environment might prevent hardlinks (e.g. SELinux restrictions in Android). As such fuzzer is failing to write identified crashes. A new copyFile helper function has been introduced to bypass such hardlink restrictions by failing over to a manual POSIX implementation.
    - Destination file O_EXCL attr. is ensuring that unique crashes feature will be functionally identical since existing files are successfully detected.

*  .gitignore update to exclude example binary

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ mach_excServer.h
 mach_excServer.c
 libs
 obj
+examples/targets/badcode1
--- a/android/Android.mk
+++ b/android/Android.mk
@@ -18,6 +18,12 @@ LOCAL_PATH := $(abspath $(call my-dir)/..)
 # Enable Linux ptrace() instead of POSIX signal interface by default 
 ANDROID_WITH_PTRACE ?= true
 
+# Make sure compiler toolchain is compatible / supported
+ifneq (,$(findstring clang,$(NDK_TOOLCHAIN)))
+  $(error Clang toolchains are not supported yet. Clang uses __aeabi_read_tp to \
+  implement thread_local, which isn't supported by bionic [$(NDK_TOOLCHAIN)])
+endif
+
 ifeq ($(ANDROID_WITH_PTRACE),true)
   ifeq ($(APP_ABI),$(filter $(APP_ABI),armeabi armeabi-v7a))
     ARCH_ABI := arm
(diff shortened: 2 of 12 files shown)


anestisb:
Two more small improvements:

* report: Save actual crash name
    - Instead of tmp runtime filename save in the report file the actual crash filename. Increases usability / scriptability of report files.

* Dynamic workspace
    - Fuzzing target environment might deploy security controls that require input files being loaded from whitelisted locations. Currently honggfuzz is using only current directory as workspace to create runtime files and save crashes. Certain setups might result into dead-end if whitelisted locations are marked as non-exec.
    - This commit allows the workspace directory to be dynamically defined with the "-W" argument. Workspace directory is used for mangled runtime files, tmp perf data & to save crashes. Report file is left untouched since it can controlled with existing "-R" flag.

    - Issue has been identified while fuzzing Android media daemons where SELinux allows input files to be loaded only from limited RO NX paths.

anestisb:
Resolved conflicts from your recent refactor changes. Also pushed a small fix for posix arch related to files_copyFile().

Unfortunately commit:https://github.com/google/honggfuzz/commit/ea39a8f5c96af12e32f79a1a95e43a92a17e3d20 has partially broken Android. Something is going wrong with thread safety locks and main exec loop is not unblocking sem_wait in case of CTRL+C SIGINT signal. Will review your refactor in fuzz.c at some point to fix this (someone need to manually kill pid to stop fuzzer in case of maxMutations not set).

robertswiecki:
I got rid of semaphores in honggfuzz, and started relying on signals/atomics exclusively. I think this will simplify things for android/maxosx

anestisb:
Yup getting rid of the semaphores solved the issues. Will resolve the conflicts and do some more testing just to be sure. Will update this pull request as soon it's done.

anestisb:
Resolved conflicts with refactored code and re-tested small fixes / improvements in this pull request. Display & refactor code appears to work fine with Android too.

robertswiecki:
As always, thank you!

PR #29, opened 2015-09-27
Stackhash signatures blacklist

anestisb:
### Description
The idea is to skip during fuzzing time already analyzed crashes without having to transfer entire crash files between target workspaces. Additionally, same vulnerable library might be loaded in different addresses (even with ASLR disabled) resulting into noise duplicates (due to PC or ADDR) that can be avoided with stack hash blacklists.

### Implementation
* New calling argument (-B) exported so that user can provide a file with blacklisted stack hashes (hex format one per line). Input file must be sorted (using the provided bash script is strongly recommended).
* File is parsed during init phase and stored in heap
* File is checked for being sorted and not empty
* When crash is detected for MAC & LINUX arch, stack hash is checked against the blacklist using a relatively fast interpolation search against the heap array.
* Stack hash blacklist is working independently of the unique crashes feature.

### Helper bash script
* tools/createStackBlacklist.sh script can be used post-campaign (after initial analysis) to extract stack hashes from crash files (following HF convention) and create a sorted blacklist file. Same script can be also used to sort existing blacklist files.
* Due to created filenames between Linux PTRACE arch and MAC not following the same name conventions, bash script takes as an argument the architecture name that fuzzer has run against. In order to be more flexible it would a good idea to consider following the same naming conventions for crashes (unique and non-unique) for both Linux & MAC.

### DIFF
--- a/common.h
+++ b/common.h
@@ -80,6 +80,9 @@ typedef struct {
     char *externalCommand;
     const char *dictionaryFile;
     const char **dictionary;
+    const char *blacklistFile;
+    uint64_t *blacklist;
+    size_t blacklistCnt;
     long tmOut;
     size_t dictionaryCnt;
     size_t mutationsMax;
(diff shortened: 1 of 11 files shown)


robertswiecki:
Nice! I'll check it out

PR #40, opened 2015-11-25
Linux stackhash improvements

anestisb:
Linux callstack hash wasn't handling the single frame unwind case in a safe manner. If only one frame was recovered (e.g. garbage pc due to stack corruption), the generated hash is not safe to be used for uniqueness decisions. Specially considering the case of a zero pc which some times might translate into 0x0 signal addr and invalid assembly instr. Ideally the unwinder needs to be improved (e.g. for ARM follow an unwind method based on LR for major corrupted frame), although let's firstly try to deal with it at the fuzzer level.

This pull request aims to deal with the single frame case by disabling unique flag for such crashes. Disabling unique flag will set a timestamp at filename resulting into always saving those crashes. Additionally, all single frame crashes stack hashes are masked with a constant prefix at the 6 MSB (BADBAD) so that post fuzzing scripts that process hashes for blacklist file can identify and skip them. Despite the duplicate, masking is performed inside "arch_hashCallstack()" procedure for cleaner code.

### DIFF
--- a/cmdline.c
+++ b/cmdline.c
@@ -174,7 +174,6 @@ bool cmdlineParse(int argc, char *argv[], honggfuzz_t * hfuzz)
 
         .timeStart = time(NULL),
         .mutationsCnt = 0,
-        .singleFrameCrashesCnt = 0,
         .crashesCnt = 0,
         .uniqueCrashesCnt = 0,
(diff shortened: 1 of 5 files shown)


robertswiecki:
Thanks for the nice set of patches

PR #44, opened 2015-12-27
Add sanitizer coverage feedback evolution support

anestisb:
## Overview
This PR introduces support for sanitizer coverage feedback by parsing sancov files generated from targets compiled with clang sanitizer (ASan, MSan, UBSan, etc.) coverage enabled. Feedback data parsing is very simple for now (only #PCs is parsed), although more advanced counters (such as BasicBlock) are on the todo list. Some additional feedback decision improvements have been introduced while working with sancov parsing.

Performance overhead for sancov enabled targets has been noticed roughly around 30-40% for some basic media file format decoders. Thus still have mixed thoughts as to what extend it's worth being used. While working with the evolutionary engine another thought that popped is to use feedback data to generate a set of test files with good coverage and then disable evolution for certain time-frames. Will do some more testing on that and follow-up if it's worth it. Feedback & request for improvements are welcome.

## Linux feedback improvements (expire dynamic file)
While working integrating sancov noticed that prepareFileDynamically is choosing an initial seed for the dynamic file and this file is never replaced. As such until campaign is stopped and restarted the same dynFile will be used. I think this is not desired for most cases since feedback evolution might hit a dead-end and thus trap in an exec loop that doesn't present any improvements or crashes. As an improvement effort in that area an additional iterations counter has been added to flag expired dynFiles. Counter is increased from main fuzzLoop and when hits  _HF_MAX_DYNFILE_ITER file is replaced with new seed. Counter is reset if unique crash is detected or better coverage is achieved. Additionally, counter has its two MSB zero-set when a non-unique crash is detected aiming for faster exist from already tested areas. Current max is 8192 and can be adjusted via common header.

## Sanitizer coverage support
sancov.c implements a basic PC enumeration from .sancov or .sancov.raw files. More feedback [data](http://clang.llvm.org/docs/SanitizerCoverage.html) parsing is on the todo list. New struct has been introduced to do the book-keeping of the counters following same pattern with HW perf.

Sanitizer flags are adjusted in Linux arch.c to export the appropriate flags required to enable sanitizer coverage data generation. sancov files are deleted after being parsed.

Sancov feedback can be enabled with "-C" flag. While sancov is developed it's prevented from being used in parallel with HW perf feedback. This can change in the future if sancov reaches a stable state.

## Generic improvements
* Some basic stack size monitor logic has been added to verify at compile time that stack sizes are maintained at accepted levels covering possible less resource-full targets such as old Android OS devices.
* Export ANDROID_NDK_TOOLCHAIN for cases were user wants to replace default Android NDK toolchain.
* Verify workDir directory exists before starting fuzzLoop
* Various typos and print format fixes
* Reduce size of display stack helper buffer

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,6 +133,7 @@ endif
 ANDROID_DEBUG_ENABLED ?= false
 ANDROID_APP_ABI       ?= armeabi-v7a
 ANDROID_API           ?= android-21
+ANDROID_NDK_TOOLCHAIN ?=
 NDK_BUILD_ARGS :=
 ifeq ($(ANDROID_DEBUG_ENABLED),true)
     NDK_BUILD_ARGS += V=1 NDK_DEBUG=1 APP_OPTIM=debug
@@ -174,7 +175,8 @@ depend:
 .PHONY: android
 android:
 	ndk-build NDK_PROJECT_PATH=. APP_BUILD_SCRIPT=./android/Android.mk \
-                  APP_PLATFORM=$(ANDROID_API) APP_ABI=$(ANDROID_APP_ABI) $(NDK_BUILD_ARGS)
+                  APP_PLATFORM=$(ANDROID_API) APP_ABI=$(ANDROID_APP_ABI) \
+                  NDK_TOOLCHAIN=$(ANDROID_NDK_TOOLCHAIN) $(NDK_BUILD_ARGS)
 
 
 # DO NOT DELETE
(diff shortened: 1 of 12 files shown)


robertswiecki:
Great set of patches! Thanks!

anestisb:
Thanks. Still working on efficient options to improve feedback engine. Longterm goal is to efficiently detect newly discovered areas in target vs maximum discovered areas in target (max counters not always the best option). Sancov seems to open some doors towards that path.

PR #45, opened 2015-12-28
DYNFILE: Fix flipRate skip bug

anestisb:
Logic has been reverted throughout the refactors leaving a wrong check behind. This PR fixes a flipRatio check resulted into wrongly skipping mangling when dynamicFile prepare method is chosen.


anestisb:
Credits to @mfence for spotting the bug

PR #56, opened 2016-02-09
LINUX: Moar remote pid enhancements

anestisb:
### Fix remote pid re-attach bug
When pid from file flag used and remote pid has been updated (new pid read from file), static boolean flag will never change preventing reapChild from attaching to new pid. Slightly modify the logic to attach again to long-lived process when new pid detected.

### Use procfs to resolve remote pid cmd
Use remote pid cmdline string for display screen and reports


robertswiecki:
Nice set of patches! Thanks.

Issue #57, opened 2016-02-10
Wiki link?


Manouchehri:
I might just be tired, but the link to the wiki/docs on line 4 of README.md seems to link back to itself?

robertswiecki:
I believe this was patched by you diff - https://github.com/google/honggfuzz/pull/58

Manouchehri:
Indeed it was. =)

PR #58, opened 2016-02-10
Link to USAGE.md instead of itself (README.md).

Manouchehri:
USAGE.md contains the features and usage details, and README.md does not.

Reference: #57.


robertswiecki:
Thanks for the fix!

PR #65, opened 2016-04-27
Small bug fixes

anestisb:
### Fix verifier null-deref bug
Verifier threads were not setting the timer resulting into a null-deref crash when first call to timer_gettime() from verifier thread.

### Fix memory leak
Detected with valgrind under the Android OS platform


robertswiecki:
Nice! Thanks.

PR #69, opened 2016-05-06
Android clang support

anestisb:
Thanks to latest Android NDK r11 current honggfuzz codebase can be cross-compiled with clang3.6 toolchains using latest API 23 (Marshmallow). With some tuning it might be possible to support API 21 & 22 too (Lollipop), although for the moment I don't thing it's worth it.

The tricky bit was libBlocksRuntime since it's not part of the NDK. The AOSP llvm compiler-rt sources haven been copied across (without any changes to the source code) to statically cross-compile BlocksRuntime (just generated an Android.mk for the lib). For the time being BlocksRuntime needs to be updated manually, although compatibility is not expected to break for same clang version. Maybe we can add some update automation later when NDK starts supporting clang 3.7

More details are available in commit logs.


robertswiecki:
Impressive work! Thanks.

PR #71, opened 2016-05-07
[MAC] Add HF report file support

anestisb:
Has the same volume of information as the Linux ptrace arch, except the crashing frame stack traces. Since on MAC that information is available from crash wrangler as a huge description string, it needs some manual (and painful) string parsing and thus is not implemented yet.

If for some reason that information is required, users can enable file logging and compile with debug flag where the `write_crash_report()` function will be activated and write entire crash dump in file log.

This PR is also removing the recently added original seed filename info in crash file name, since it creates very large names are not very practical and might get truncated due to size limits. That information is available in the report now.

### DIFF
--- a/mac/arch.c
+++ b/mac/arch.c
@@ -444,10 +444,11 @@ bool arch_archInit(honggfuzz_t * hfuzz)
     return true;
 }
 
+#ifdef DEBUG
 /*
  * Write the crash report to DEBUG
  */
-void
+static void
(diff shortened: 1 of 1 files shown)


tl0gic:
@anestisb let me run some tests with this PR and I will get back to you.

tl0gic:
@anestisb I tested this PR, it works well and has all the necessary information.
I will spend sometime this week to add the stack traces as in the linux reports.

PR #78, opened 2016-07-21
Update sancov.c

bryongloden:
[[../honggfuzz-master/sancov.c:454](https://github.com/google/honggfuzz/blob/master/sancov.c#L454)]: (error) Common realloc mistake: 'mapsBuf' nulled but not freed upon failure

**[This solution](http://stackoverflow.com/a/1607031) from Stack Overflow was helpful in fixing the error because if realloc cannot find enough space, it returns a null pointer, and leaves the previous region allocated.** 

Found by https://github.com/bryongloden/cppcheck


bryongloden:
I signed it!

robertswiecki:
Thanks, good catch. I decided to solve it everywhere :) https://github.com/google/honggfuzz/commit/f49962d0c846a94fd446e679fbf18247b9efe92d

bryongloden:
LGTM ;)

PR #79, opened 2016-07-26
Update files.c

bryongloden:
Checking [../honggfuzz-master/files.c](https://github.com/google/honggfuzz/blob/master/files.c): `__clang__`...
[[../honggfuzz-master/files.c:144](https://github.com/google/honggfuzz/blob/master/files.c#L144)]: (error) Memory leak: `buf`
[[../honggfuzz-master/files.c:371](https://github.com/google/honggfuzz/blob/master/files.c#L371)]: (error) Memory leak: `inFileBuf`

Found by https://github.com/bryongloden/cppcheck

### DIFF
--- a/files.c
+++ b/files.c
@@ -141,6 +141,7 @@ bool files_writePatternToFd(int fd, off_t size, unsigned char p)
     memset(buf, p, (size_t) size);
     int ret = files_writeToFd(fd, buf, size);
 
+    free(buf);
     return ret;
 }
 
@@ -368,6 +369,7 @@ bool files_copyFile(const char *source, const char *destination, bool * dstExist
     ssize_t readSz = files_readFromFd(inFD, inFileBuf, (size_t) inSt.st_size);
     if (readSz < 0) {
         PLOG_E("Couldn't read '%s' to a buf", source);
+        free(inFileBuf);
         return false;
     }
 


robertswiecki:
Probably too many intermediate changes to merge it now. I'll go ahead and take a look individually for those reports.

bryongloden:
We can sync with the upstream master then reapply our patch.
At that point merge in our fix so we get credit for authoring.
Sound okay @robertswiecki?

anestisb:
I think that cppcheck is not picking the defer blocks thus missing the free() calls triggered when function loses scope. For both reported cases the code inside the defer block will trigger the mentioned heap buffer frees. As such, I believe that they're both false positives.

I've run with valgrind under Android and haven't noticed any reported leaks. Do you have any different evidence from any memory inspection tool? If such, it means that defer blocks are not working as expected under certain cases / setups?

bryongloden:
The `defer` construct is new to us, but if it's _triggered_ after losing scope, it's logical that memory will be freed in both cases. We've had to adopt some unsound techniques in our tool because sound and complete static analysis is shown to be undecidable. Notwithstanding the logic of `defer`, manual inspection should have detected this false positive. Our apology for not catching that. Valgrind is a dynamic analysis tool, cppcheck is not. No other security testing tools were used for this bug scan.

robertswiecki:
So... I guess we can close it?:)

PR #88, opened 2016-09-10
display alignment

riusksk:
fix the bug that liblzma library symbols not found in ubuntu, and display alignment.

Before:
```
=============================| honggfuzz v.0.8rc |============================
Iterations: 63862 [63.86k]
Run time: 15 hrs 22 min 7 sec (since: 2016-09-09 23:19:59)
Input file/dir: '/Users/riusksk/Downloads/mp4'
Fuzzed cmd: '/Applications/test.app/Contents/MacOS/test ___FILE___'
Fuzzing threads: 5, CPUs: 4, CPU: nan% (nan%/CPU)
Execs per second: 0 (avg: 1)
Crashes: 8 (unique: 3, blacklist: 0, verified: 0)
Timeouts (4 sec): 0
==================================| LOGS |====================================
```

After:
```
-----------------------------[ honggfuzz v0.8rc ]-----------------------------
      Iterations : 5
        Run Time : 0 hrs 0 min 3 sec (since: 2016-09-10 16:00:17)
  Input File/Dir : '/Users/riusksk/Downloads/mp4'
      Fuzzed Cmd : '/Applications/test.app/Contents/MacOS/test ___FILE___'
 Fuzzing Threads : 5, CPUs: 4, CPU: nan% (nan%/CPU)
     Execs Speed : 0/sec (avg: 1)
         Crashes : 0 (unique: 0, blacklist: 0, verified: 0)
Timeouts (4 sec) : 0
-----------------------------------[ LOGS ]-----------------------------------
```


robertswiecki:
Thanks, I like the display patch.

Could you provide mroe data on the lzma problem. I believe that when compiling dynamically the libraries should be pulled in by another libraries (using DT_NEEDED segments in ELF). We don't use lzma directly, so I'd like to reproduce the problem.

Also, one of your pull reqs contains some mp4 files. You removed them in a subsequent commit, but still they are there, as in they will be updated to the git base if I merged the pull requests as is. Probably cleaning it a bit would help?

riusksk:
the liblzma bug only in some ubuntu env, beacause ld error when compiling libunwind, it can't found some liblzma functions symbols, for example "Symbol not found: _lzma_auto_XXX".

add ' -lzma' can to avoid the above problems, but this bug only appeared  on my an ubuntu virtual.

some mp4 files are test file, you can ignore.

riusksk:
How can I remove the commits about mp4 files?

riusksk:
update ubuntu can solve it

PR #89, opened 2016-09-10
display alignment and fix liblzma bug

riusksk:
fix the bug that liblzma library symbols not found in ubuntu, and display alignment.

the liblzma bug only in some ubuntu env, beacause ld error when compiling libunwind, it can't found some liblzma functions symbols, for example "Symbol not found: _lzma_auto_XXX".
add `-lzma` can to avoid the above problems, but this bug only appeared on my an ubuntu virtual.

Before:
```
=============================| honggfuzz v.0.8rc |============================
Iterations: 63862 [63.86k]
Run time: 15 hrs 22 min 7 sec (since: 2016-09-09 23:19:59)
Input file/dir: '/Users/riusksk/Downloads/mp4'
Fuzzed cmd: '/Applications/test.app/Contents/MacOS/test ___FILE___'
Fuzzing threads: 5, CPUs: 4, CPU: nan% (nan%/CPU)
Execs per second: 0 (avg: 1)
Crashes: 8 (unique: 3, blacklist: 0, verified: 0)
Timeouts (4 sec): 0
==================================| LOGS |====================================
```

After:
```
-----------------------------[ honggfuzz v0.8rc ]-----------------------------
      Iterations : 5
        Run Time : 0 hrs 0 min 3 sec (since: 2016-09-10 16:00:17)
  Input File/Dir : '/Users/riusksk/Downloads/mp4'
      Fuzzed Cmd : '/Applications/test.app/Contents/MacOS/test ___FILE___'
 Fuzzing Threads : 5, CPUs: 4, CPU: nan% (nan%/CPU)
     Execs Speed : 0/sec (avg: 1)
         Crashes : 0 (unique: 0, blacklist: 0, verified: 0)
Timeouts (4 sec) : 0
-----------------------------------[ LOGS ]-----------------------------------
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ ifeq ($(OS),Linux)
                    -Wno-unknown-warning-option -funroll-loops \
                    -D_FILE_OFFSET_BITS=64
     ARCH_LDFLAGS := -L/usr/local/include -L/usr/include \
-                    -lpthread -lunwind-ptrace -lunwind-generic -lbfd -lopcodes -lrt
+                    -lpthread -lunwind-ptrace -llzma -lunwind-generic -lbfd -lopcodes -lrt
     ARCH_SRCS := $(wildcard linux/*.c)
 
     ifeq ("$(wildcard /usr/include/bfd.h)","")
--- a/display.c
+++ b/display.c
@@ -140,7 +140,7 @@ static void display_displayLocked(honggfuzz_t * hfuzz)
     display_put("%s", ESC_CLEAR);
     display_put("-----------------------------[ %s v%s ]-----------------------------\n", PROG_NAME,
                 PROG_VERSION);
-    display_put("Iterations: " ESC_BOLD "%" _HF_MONETARY_MOD "zu" ESC_RESET, curr_exec_cnt);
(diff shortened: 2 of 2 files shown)


robertswiecki:
Thanks... so I'll take the display patch which is awseome.

But, I don't think I'll add -llzma to Makefile. I know that it might fix some problem, but I personaly think that the problem lies somewhere else.

We don't use lzma directly, and all libraries needing it (like unwind-ptrace) should depend on it, like here.

$ ldd /usr/lib/x86_64-linux-gnu/libunwind-ptrace.so.0
	...
	liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5 (0x00007f5fed86c000)
        ...


If your library don't depend on it, then it's a bug, and I think the library compilation for your system should be fixed. Otherwise we'll be linking (and requiring) in lzma always, no matter whether it's needed or not.

Would you be willing to investigae why/if your libunwid is compiled correctly, and eventually asking Ubuntu maintainers for a fix (if needed)?

robertswiecki:
I merged it, and later removed -llzma, it in turn makes problem with compiling on my machine, probabl;y because I don't have liblzma-dev installed. I still think this needs to be fixed, and if you need it for compiling on your machine, you can always:

LDFLAGS="-llzma" make

PR #97, opened 2016-10-14
 Updated Makefile (on OSX) to use xcrun (from xcode) versus sw_vers t…

aaronc100:
…o discover SDK version. This prevents failed builds when using OSX 10.11 with the newest Xcode 10.12.


aaronc100:
CLA submitted.

tl0gic:
I think this solution is not going to work in all instances.

OS_VERSION is used to determine which CRASH_REPORT to link into your binary. If you use the wrong one it won't build a proper binary for your OS X version.

What you need to change is SDK_NAME and for that I think it is a good idea to use xcrun --show-sdk-version instead of hardcoding the name as we do now.

PR #98, opened 2016-10-14
Updated Makefile (on OSX) to use xcrun (from xcode) versus sw_vers to…

aaronc100:
… discover SDK version. This prevents failed builds when using OSX 10.11 with the newest Xcode 10.12.


tl0gic:
xcrun doesn't give you the OS_VERSION. It is used to get the SDK version.

You should change SDK_NAME instead of OS_VERSION. Otherwise, if Apple releases a new CRASH_REPORTER your build will break.

PR #100, opened 2016-10-15
Updated the darwin area of the Makefile to more accurate variable nam…

aaronc100:
…es: changed OS_VERSION to OSX_SDK_VERSION.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,10 +72,10 @@ ifeq ($(OS),Linux)
     # OS Linux
 else ifeq ($(OS),Darwin)
     ARCH := DARWIN
+
+    # Figure out which crash reporter to use.
     CRASHWRANGLER := third_party/mac
     OS_VERSION := $(shell sw_vers -productVersion)
-    OSX_SDK_VERSION := $(shell xcrun --show-sdk-version)
-
 	ifneq (,$(findstring 10.12,$(OS_VERSION)))
 		CRASH_REPORT := $(CRASHWRANGLER)/CrashReport_Yosemite.o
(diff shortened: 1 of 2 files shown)


aaronc100:
Most of these changes are based upon Haris Andrianakis' comments to my last PR.  Thanks!

robertswiecki:
Hi @tl0gic - will you take this - I've already too much mess to the macos :)

Makefile
```diff
@@ -73,26 +73,39 @@ ifeq ($(OS),Linux)
...
+        CRASH_REPORT := $(CRASHWRANGLER)/CrashReport_Mountain_Lion.o
+    else
+        $(error Unsupported MAC OS X version)
+    endif
+
+	ifneq (,$(findstring 10.12,$(OSX_SDK_VERSION)))
+		SDK_NAME := "macosx10.12"
     else ifneq (,$(findstring 10.11,$(OSX_SDK_VERSION)))
         # El Capitan didn't break compatibility
         SDK_NAME := "macosx10.11"
```
tl0gic: I think we can avoid the if/else conditions here if we do something like this:
SDK_NAME := macosx$(OSX_SDK_VERSION)

What do you think?

tl0gic:
Hey @robertswiecki, sure thing!

@aaronc100 thanks for doing this! I think overall it looks good, I left you one small comment. Let me know what you think.

Makefile
aaronc100: Yeah, def, let me give that a shot!

PR #101, opened 2016-10-16
keep fuzzed file orig file ext

riusksk:
it can fuzzing the mulitple file format in the same time, for example, some image file(*.png, *.gif, *.jpg...) in the following file dir:

`honggfuzz -f file -e any -- imageview ___FILE___`

Some app must only parse the specific file suffix name.,but the default generated file suffix name is *.fuzz, and -e arg can only specify a file suffix, so the fuzzing would be invalid when input dir have multiple file suffix name.

### DIFF
--- a/common.h
+++ b/common.h
@@ -289,7 +289,7 @@ typedef struct {
     const char *origFileName;
     char fileName[PATH_MAX];
     char crashFileName[PATH_MAX];
-    char ext[PATH_MAX];
+    char ext[NAME_MAX];
     uint64_t pc;
     uint64_t backtrace;
     uint64_t access;
--- a/common.h
+++ b/common.h
@@ -289,6 +289,7 @@ typedef struct {
     const char *origFileName;
     char fileName[PATH_MAX];
     char crashFileName[PATH_MAX];
+    char ext[PATH_MAX];
     uint64_t pc;
     uint64_t backtrace;
     uint64_t access;
--- a/fuzz.c
+++ b/fuzz.c
@@ -508,8 +508,9 @@ static void fuzz_fuzzLoop(honggfuzz_t * hfuzz, fuzzer_t * fuzzer)
 	//printf("ext: %s\n",fuzzer->ext);
 	
     }
-    
-    hfuzz->fileExtn = fuzzer->ext;
+    if(!strcmp(hfuzz->fileExtn, "*")){ 
+    	hfuzz->fileExtn = fuzzer->ext;
+    }
     fuzz_getFileName(hfuzz, fuzzer->fileName);
 
     if (state == _HF_STATE_DYNAMIC_PRE) {
--- a/fuzz.c
+++ b/fuzz.c
@@ -63,11 +63,6 @@ static void fuzz_getFileName(honggfuzz_t * hfuzz, char *fileName)
              (int)getpid(), (unsigned long int)tv.tv_sec, util_rnd64(), hfuzz->fileExtn);
 }
 
-void fuzz_getExtension(const char *file_name,char *extension)  
-{  
-    extension = strrchr(file_name, '.' );   
-}
-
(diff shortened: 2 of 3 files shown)


anestisb:
I don't think that all this overhead is necessary. If I'm not mistaken the fuzzed file extension has not been added with the concept of preserving the original extension. It's just a marker to separate mutated files from input corpus (keep in mind that not all runs have a meaningful file extension - e.g. external tools, stdin, etc.). To put it simply, `.fuzz` (or any other selected by the user) is an extension to identify files generated from honggfuzz regardless of the selected corpus.

Modifying the output extension will also break most post-fuzzing automation tools that rely on a constant suffix to pick-up fuzzer output crashes or dynamically generated files.

I don't see any practical reason for this overhead, since the generated fuzzer report includes the original seed file the mutation of which resulted into the matching crash.

So is there effectively any reason to dynamically modify the extension of output files?

anestisb:
Also forgot to mention that in case the fuzzing target expects the input files to be of a specific extension, you can utilize `--pprocess_cmd` option to rename files based post mangling and before running worker.

riusksk:
I know what you mean, so keep all original functions, only add "-e any" arg option to fuzz multiple file formats for some GUI app which requires correct ext name. To improve the fuzzing effect and avoid only fuzzing one file format for gui app. For example, preview (image view er) in mac os.

fuzz.c
```diff
@@ -63,6 +63,22 @@ static void fuzz_getFileName(honggfuzz_t * hfuzz, char *fileName)
              (int)getpid(), (unsigned long int)tv.tv_sec, util_rnd64(), hfuzz->fileExtn);
 }
 
+void fuzz_getExtension(const char *file_name,char *extension)  
+{  
+    int i=0,length;  
+    length=strlen(file_name);  
+        while(file_name[i])
```
anestisb: You don't need to loop around the entire string. An `strrchr` call against the dot char `.` will be enough to have the minimum perf overhead.

common.h
```diff
@@ -289,6 +289,7 @@ typedef struct {
     const char *origFileName;
     char fileName[PATH_MAX];
     char crashFileName[PATH_MAX];
+    char ext[PATH_MAX];
```
anestisb: That can be `char ext[NAME_MAX];` instead (for less size overhead) since the extension cannot be larger than the allowed max file name.

anestisb:
I personally don't mix formats under same campaign when using honggfuzz since you don't get the best out of supported instrumentation backends.

Maybe another idea would be to differentiate the file extension of the intermediate files that are fed to the worker execve, from the file extension of the output crash files that will feed all post-campaign automation.

It's mostly politics, although it doesn't feel good to add this performance overhead for every file selection iteration.

fuzz.c
```diff
@@ -65,18 +65,7 @@ static void fuzz_getFileName(honggfuzz_t * hfuzz, char *fileName)
...
-    {  
-        if(file_name[i]=='.')  
-        break;  
-        i++;  
-    }  
-    if(i<length)  
-    strcpy(extension,file_name+i+1);  
-    else  
-    strcpy(extension,"\0");  
+    extension = strrchr(file_name, '.' );
```
robertswiecki: You are updating a copy of local variable - this will not have any effect IMO.

fuzz.c
```diff
@@ -63,6 +63,11 @@ static void fuzz_getFileName(honggfuzz_t * hfuzz, char *fileName)
              (int)getpid(), (unsigned long int)tv.tv_sec, util_rnd64(), hfuzz->fileExtn);
 }
 
+void fuzz_getExtension(const char *file_name,char *extension)  
+{  
+    extension = strrchr(file_name, '.' );
```
robertswiecki: This will simply update a local copy of the 'extension' var, you probably want to pass char **extension and do

*extension = 

or simply return char*.

PR #129, opened 2017-03-25
FIX: Couldn't write buffer to file [DYNAMIC]

n30m1nd:
Each thread should have its own `[DYNAMIC#N]` file to write the buffer to.


n30m1nd:
Signed.

robertswiecki:
Thanks for the report and for the fix. I don't mind pulling it, but I'm wondering when this error fires. The fileName should not be used anywhere in the dynamic mode, except for being used in some LOG_I/LOG_D... strange.

Nevertheless, pulled.

n30m1nd:
For further triage if interested - this is the command line that on my laptop (Debian 4.9.0-0.bpo.2-rt-amd64, 16gb, i7-6700) made it crash:
`sudo ./honggfuzz --linux_perf_instr -f ~/Testcases/archives/ -q -n 8 -- /usr/bin/7z l ___FILE___`

See attached file:
image

robertswiecki:
Ah.. thanks. I think the hfuzz.fileName setting logic was flawed. I think I fixed it now with 34fe7acb7646957add576eecee8da6fc3ea7f1c9

PR #141, opened 2017-05-21
Fix paths for examples

dyjakan:
Update for outdated paths.

Also, I think that at some point loading corpus files for external fuzzers have changed. For version <= 0.7 you can provide argument `-f inputfiles/badcode1.txt` however later versions break on this (but work for just the directory).


robertswiecki:
Thanks!

PR #144, opened 2017-05-29
sort input files

bmwiedemann:
when building packages (e.g. for openSUSE Linux)
(random) filesystem order of input files
influences ordering of functions in the output,
thus without the patch, builds (in disposable VMs) would always differ.

See https://reproducible-builds.org/ for why this matters.


bmwiedemann:
I signed it! (now)

robertswiecki:
Thanks for the patch!

PR #166, opened 2017-11-02
Some links fixes, added or updated.

kant:


robertswiecki:
Thanks! This was very much needed.

PR #167, opened 2017-11-03
Add macOS 10.13 (High Sierra) Support

anestisb:
* Updated crash wrangler to work with the latest Xcode under 10.13
* Fixed `MSG_NOSIGNAL` flag compatibility


anestisb:
Btw, forgot to mention that the crashwrangler binary has been download from the official Apple [website](https://developer.apple.com/library/content/technotes/tn2334/_index.html).

PR #171, opened 2017-12-04
android: NDK r16 compatibility & switch default to clang

anestisb:
Android NDK r16 deprecates (although doesn't remove yet) the gcc 4.9 support.
However, some toolchain flags have been altered thus need the
`ANDROID_NDK_TOOLCHAIN_VER` flag to be also set to effectively select gcc
over default clang.

This commit is also switching the default compiler selection from gcc to clang. It
has been tested the past few months and no issues surfaced so far against
various aarch64 Android targets.


robertswiecki:
Nice, thanks!

PR #172, opened 2017-12-12
Fix parameter type for clang

zachriggle:
Clang expects a uintptr_t and fails fatally:

    fatal error: error in backend: Sanitizer interface function redefined: void (i32)* bitcast (void (i8*)* @__sanitizer_cov_trace_pc_indir to void (i32)*)


robertswiecki:
Sorry, didn't notice this before the internal conversation took place. Fixed in https://github.com/google/honggfuzz/commit/6e1ab71523e995d15db5980b3d6677168468bab1

PR #174, opened 2017-12-13
Fix issues on Android system builds

zachriggle:
Bionic's implementation of strcmp() is hijacked by Honggfuzz, and invoked before
either Address Sanitizer or the trace-pc-guard coverage initialization routines
have been invoked.

This leads to segfaults on Android builds with ASAN+coverage.

Use trace_pc_guard as a convenient place to check for ASAN's initialization.
Defer trace_pc_guard activity until coverage is initialized.

Change-Id: I1dfa1946b8baad1ac58f39917d59b74ffd25bdd4

### DIFF
--- a/libhfuzz/instrument.c
+++ b/libhfuzz/instrument.c
@@ -21,6 +21,7 @@
 #include "libcommon/util.h"
 
 int hfuzz_module_instrument = 0;
+static bool inited = false;
 
 /*
  * We require SSE4.2 with x86-(32|64) for the 'popcnt', as it's much faster than the software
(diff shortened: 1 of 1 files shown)


robertswiecki:
Though, I will probably later surround this by HF_ANDROID ifdefs, cause it's always one more cmp per block.

robertswiecki:
I hope it'll work with https://github.com/google/honggfuzz/commit/52815ae3722d4028c3c0a32d5f892a6dde40cdbb as this check might be costly for other platforms (which hopefully don't instrument their ld.so)

PR #184, opened 2018-01-21
Support external fuzzer and provide code coverage

dobin:
Aka The FFW honggfuzz patch.

This proposed patch for honggfuzz is used by the fuzzing for worms (FFW) framework available here: github.com/dobin/ffw
This patch makes it possible to use honggfuzz as a code coverage tool for long-living (network-) servers.

It provides the following functionality:
* A socket interface for an external fuzzer
* Support for code coverage of long living processes

The socket interface supports the following messages:
* Honggfuzz -> Fuzzer: "Fuzz": Send fuzz data to target
* Honggfuzz -> Fuzzer: "Cras": Target crashed
* Honggfuzz -> Fuzzer: "New!": New basic block reached
* Fuzzer -> Honggfuzz: "okay": Fuzzer sent his stuff to target
* Fuzzer -> Honggfuzz: "bad!": The server is unreachable

Honggfuzz will implement the following tasks:
* Start the target
* Restart the target if it crashes
* Restart the target if a crash/hang is detected by the fuzzer
* Provide code coverage information about new basic blocks

An example vulnerable server plus tests have been provided in the
`socketfuzzer/` directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -356,7 +356,7 @@ android-clean-deps:
 	done
 
 # DO NOT DELETE
-
+socketfuzzer.o: socketfuzzer.h
 cmdline.o: cmdline.h honggfuzz.h libhfcommon/util.h libhfcommon/common.h
 cmdline.o: libhfcommon/files.h libhfcommon/common.h libhfcommon/log.h
 display.o: display.h honggfuzz.h libhfcommon/util.h libhfcommon/common.h
--- a/cmdline.c
+++ b/cmdline.c
@@ -304,6 +304,8 @@ bool cmdlineParse(int argc, char* argv[], honggfuzz_t* hfuzz) {
 
         .feedback_mutex = PTHREAD_MUTEX_INITIALIZER,
 
+        .socketFuzzer = false,
+
         .cnts =
             {
                 .mutationsCnt = 0,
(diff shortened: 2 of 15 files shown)


dobin:
Main problems which i tried to solve in the honggfuzz codebase: 
* Identify when a process is crashed
* Identify when a process needs to be restarted

Shortcomings: 
* The unit test socketfuzzer/unittest.sh will fail regularly and randomly. It appears like the new basic block identification is not reliable (HF bug?)
* Identification of crashes in HF is performed in trace.c and arch.c. There should be an easier, reliable way?
* had to introduce a variable "hasCrashed", to indicate that the long living process is crashed (either detected by HF, or by the fuzzer / externally)
* The unit test will fail  if the binary is compiled with HFUZZ_CC_ASAN
* I cannot compile the unit tst atm with hfuzz-clang, only hfuzz-gcc

While this patch works for me, i dont feel the quality is adequate to integrate it into honggfuzz. Feedback is welcome.

dobin:
I signed it!

robertswiecki:
I guess later I'll do some small naming convention unifications, but for now, let's pull it in! :)

robertswiecki:
Nice set of changes! Thanks for working on that. 

PS: I've done a quick make android (some format string fixes: https://github.com/google/honggfuzz/commit/f3ecc0ede2ec12344045324de73165ef90a9c8ac) and later make indent depend (https://github.com/google/honggfuzz/commit/562761928d542c6559ae89f7c9500bbda45a0184)

dobin:
Nice, thanks a lot! I'll follow up with more tests / feedback.

PR #204, opened 2018-05-05
Fix IPT error reporting

q3k:
The errorcodes returned by libipt functions are returned as negative numbers. We need to flip them back to positive before passing them to `pt_errstr` (otherwise we get '`internal error`' every time).


robertswiecki:
senk you sir

PR #212, opened 2018-06-29
Fix compatibility issues with NetBSD.

plusun:
1. Using -pthread instead of -lpthread flag
2. Add install option for make
3. Convert char type to unsigned char type before calling isdigit(3)
and tolower(3). Because according to the ctype(3) man page of NetBSD,
the parameter of these functions should be representable as unsigned
char to avoid undefined behaviors. For more details, please refer to:
http://netbsd.gw.com/cgi-bin/man-cgi?ctype+3+NetBSD-current

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,7 +145,7 @@ else
                    -Wextra -Wno-initializer-overrides -Wno-override-init \
                    -Wno-unknown-warning-option -Wno-unknown-pragmas \
                    -funroll-loops
-    ARCH_LDFLAGS := -lpthread -L/usr/local/lib
+    ARCH_LDFLAGS := -pthread -L/usr/local/lib
     # OS Posix
 endif
 
@@ -440,3 +440,15 @@ mac/arch.o: sancov.h subproc.h
 posix/arch.o: arch.h honggfuzz.h libhfcommon/util.h fuzz.h
 posix/arch.o: libhfcommon/common.h libhfcommon/files.h libhfcommon/common.h
 posix/arch.o: libhfcommon/log.h sancov.h subproc.h
+
+PREFIX		?= /usr/local
+BIN_PATH	=$(PREFIX)/bin
+
+install: all
+	mkdir -p -m 755 $${DESTDIR}$(BIN_PATH)
+	install -m 755 honggfuzz $${DESTDIR}$(BIN_PATH)
+	install -m 755 hfuzz_cc/hfuzz-cc $${DESTDIR}$(BIN_PATH)
+	install -m 755 hfuzz_cc/hfuzz-clang $${DESTDIR}$(BIN_PATH)
+	install -m 755 hfuzz_cc/hfuzz-clang++ $${DESTDIR}$(BIN_PATH)
+	install -m 755 hfuzz_cc/hfuzz-gcc $${DESTDIR}$(BIN_PATH)
+	install -m 755 hfuzz_cc/hfuzz-g++ $${DESTDIR}$(BIN_PATH)
--- a/libhfcommon/util.c
+++ b/libhfcommon/util.c
@@ -284,11 +284,11 @@ int64_t fastArray64Search(uint64_t* array, size_t arraySz, uint64_t key) {
 }
 
 bool util_isANumber(const char* s) {
-    if (!isdigit(s[0])) {
+    if (!isdigit((unsigned char)s[0])) {
         return false;
     }
     for (int i = 0; s[i]; s++) {
-        if (!isdigit(s[i]) && s[i] != 'x') {
+        if (!isdigit((unsigned char)s[i]) && s[i] != 'x') {
             return false;
         }
     }
--- a/libhfuzz/memorycmp.c
+++ b/libhfuzz/memorycmp.c
@@ -44,14 +44,14 @@ static inline int _strcasecmp(const char* s1, const char* s2, uintptr_t addr) {
     unsigned int v = 0;
 
     size_t i;
-    for (i = 0; tolower(s1[i]) == tolower(s2[i]); i++) {
+    for (i = 0; tolower((unsigned char)s1[i]) == tolower((unsigned char)s2[i]); i++) {
         if (s1[i] == '\0' || s2[i] == '\0') {
             break;
         }
         v++;
     }
     instrumentUpdateCmpMap(addr, v);
-    return (tolower(s1[i]) - tolower(s2[i]));
+    return (tolower((unsigned char)s1[i]) - tolower((unsigned char)s2[i]));
 }
 
 static inline int _strncmp(const char* s1, const char* s2, size_t n, uintptr_t addr) {
@@ -86,8 +86,8 @@ static inline int _strncasecmp(const char* s1, const char* s2, size_t n, uintptr
     int ret = 0;
 
     for (size_t i = 0; i < n; i++) {
-        if (tolower(s1[i]) != tolower(s2[i])) {
-            ret = ret ? ret : (tolower(s1[i]) - tolower(s2[i]));
+        if (tolower((unsigned char)s1[i]) != tolower((unsigned char)s2[i])) {
+            ret = ret ? ret : (tolower((unsigned char)s1[i]) - tolower((unsigned char)s2[i]));
         } else {
             v++;
         }


plusun:
I signed it!

krytarowski:
From a NetBSD developer point of view, this patch looks correct.

There is a nit that some symbols are called `_symbol`, it's a reserved namespace in the NetBSD libc and they will clash.. like `_strcasecmp`.

robertswiecki:
Sorry for late response, was recently on a short vac.

robertswiecki:
Yeah, I can change _strcasecmp and alike to something else as its a static/local func

krytarowski:
> Yeah, I can change _strcasecmp and alike to something else as its a static/local func

Please apply this change to all _symbols.

robertswiecki:
FYI: renaming done in https://github.com/google/honggfuzz/commit/d94e3e8b13f0f90d030288727750de258d93d64c

krytarowski:
Perfect!

While there, are there some tests for the ptrace(2) attach functionality? I intend to port it to NetBSD, it would be nice to have some aid to know whether it works as expected with feature parity with Linux.

We have got an odd use-case of fuzzing the ping(8) program.. maybe there would be an option to attach to a running ping(8) program and feed it with randomized ICMP content.

So far traditional integration of AFL, honggfuzz, libFuzzer and radamsa over some C API failed to get integration without massive code rewrite.

robertswiecki:
Hm.. the ptrace code is quite Linux specific, even assuming some emulation in NetBSD (I guess it's similar to FreeBSD's?).

There are no automatic tests, I've been testing it manually with just few fuzzing setups I have (under Linux, MacOS and with FreeBSD).

As for ptrac'eing data into ping and other apps, this sounds viable technically, but you'd have to know the addresses, and run the app without ASLR, so not sure how many people would use this mode. Maybe it'd be more useful to create some LD_PRELOAD lib for ping which hijacks sendmsg or whatever is used with ping, and fills the buffer with fuzzed data?

robertswiecki:
In case you use the LD_PRELOAD trick, you can use HF_ITER() to fetch the data from the fuzzer (it works the opposite way to LLVMFuzzerTestOneInput which is called, as opposed to HF_ITER() which must be called)

krytarowski:
FYI: honggfuzz/NetBSD mentioned in the GSoC report:

GSoC 2018 Reports: Integrate libFuzzer with the Basesystem, Part 2
http://blog.netbsd.org/tnf/entry/gsoc_2018_reports_integrate_libfuzzer1

robertswiecki:
Ha.. a really solid work! Thanks for making it available.

plusun:
Hi, I've tried to use LD_PRELOAD method together with HF_ITER interface. But I encountered some problems.

Here is my source code: https://github.com/plusun/src/tree/LD_PRELOAD/sbin/ping

The most important one is the **fuzzer_ops.c**, within which I shadowed some interfaces like **socket**, **poll** and etc. The most important one is **recvfrom**. They will call **HF_ITER** to fetch data and organize them into a structure.

With this setting, I compiled it into a library like this:

> hfuzz-clang -fPIC  -Wall -Werror -g -c fuzzer_ops.c
>clang -shared -fPIC fuzzer_ops.o /path/to/libhfuzz.a -o libfuzzerops.so

And then recompiled the ping(8) from source with **hfuzz-clang** and **hf-clang++**. But unfortunately, when I fuzz it with this command:

> LD_PRELOAD=$PWD/libfuzzerops.so honggfuzz -P -f input/ -- ./ping -c 1 127.0.0.1

it got error like this:

> Entering phase 1/2: Dry Run
> Launched new fuzzing thread, no. #0
> Launched new fuzzing thread, no. #1
> Launched new fuzzing thread, no. #2
> Launched new fuzzing thread, no. #3
> Launched new fuzzing thread, no. #5
> Launched new fuzzing thread, no. #4
> Launched new fuzzing thread, no. #6
> Persistent mode: Launched new persistent PID: 19225
> Launched new fuzzing thread, no. #7
> [2018-07-18T12:53:08+0000][F][18913] HonggfuzzFetchData():75 readFromFd(rcvLen, size=8) failed

There was an initial file containing a "hello, world" string under the **input** directory.

Could you please tell me what is happening with this error?

Thanks,
Yang Zheng

robertswiecki:
Do you have any warning before that, the readFromFd() should throw sth?

In any case, maybe your ping binary closes filedescriptors above some value (*BSDs have closefrom()), and honggfuzz uses 1021-1024, given that ping is set-uid?

krytarowski:
@robertswiecki btw. I might have a list of research projects related to your expertise to propose to students.. if it is viable, I can drop a mail and show the list if you are interested to have a look.

robertswiecki:
@krytarowski absolutely, please feel free to use robert@swiecki.net

plusun:
@robertswiecki Thanks for the hint, I originally thought that it was because I didn't shadow close for the socket interfaces. However, when I shadowed the "close" like others (such as "socket", "recvfrom"), the problem remained.

There seemed to be no other warnings/errors before that message. For the full output, you can refer to this file: https://github.com/plusun/src/blob/LD_PRELOAD/sbin/ping/output

robertswiecki:
Maybe this https://github.com/google/honggfuzz/commit/174b2ab951a5c7ab61068977735f9ccccae23995  will reveal some more?

plusun:
[2018-07-25T11:55:22+0000][F][13581] HonggfuzzFetchData():76 readFromFd(rcvLen, size=8) failed, received=-1 bytes

krytarowski:
Is there some errno set after the call `files_readFromFd()`?

krytarowski:
Something along these lines:

```
    uint64_t rcvLen;
    errno = 0;
    ssize_t sz = files_readFromFd(_HF_PERSISTENT_FD, (uint8_t*)&rcvLen, sizeof(rcvLen));
    if (sz != sizeof(rcvLen)) {
        LOG_F("readFromFd(rcvLen, size=%zu) failed, received=%zd bytes, errno=%d", sizeof(rcvLen), sz, errno);
    }
```

krytarowski:
Does it work without a sanitizer? Sanitizers open a lot of FDs on startup.

Please try to record execution of the process with `ktruss -i -o log.dat $PROGRAM_HERE`.

Probably we might want to wrap `LD_PRELOAD=$PWD/libfuzzerops.so honggfuzz -P -f input/ -- ./ping -c 1 127.0.0.1` with a shell script to not overwrite dynamic symbols in ktruss(1).

After that, please share `kdump log.dat`.

robertswiecki:
honggfuzz -E LD_PRELOAD=$PWD/libfuzzerops.so ...

should work as well

plusun:
Thanks a lot! "-E" flag works well!

krytarowski:
Does it mean that it's fuzzing now?

plusun:
@krytarowski Yes! I'm fuzzing it in the background now!

krytarowski:
Perfect, probably we have overloaded symbols in hongfuzz too.

robertswiecki:
Good luck!

BTW, is there any modern LiveCD for NetBSD? I guess I could use it along Linux/CygWin/FreeBSD/MacOS when releasing new versions for release testing.

robertswiecki:
Cloud image or something similar will also do

krytarowski:
This is a really great to know that LD_PRELOAD fuzzing works, I was thinkering how to perform something like fuzzing from the kernel syscall layer (like using PT_SYSCALL / PT_SYSCALLEMU) and get rid of ifdefs. I'm happy that we have managed to get a functional solution!

There is a GCE support done by @bsiegert.
Another option pmci: https://github.com/billziss-gh/pmci

There are some downstream relases as live-cd, but I propose to just install the pristine one in a VM:
http://ftp.netbsd.org/pub/NetBSD/NetBSD-8.0/images/NetBSD-8.0-amd64.iso

Proposed steps:
 - install the distribution including all sets
 - fetch pkgsrc from source
 - install security/mozilla-rootcerts (cd pkgsrc/security/mozilla-rootcerts && make install)
 - and follow post-install notes of mozilla-rootcerts
 - install git (devel/git): cd pkgsrc/devel/git && make install
 - install other programs, possibly: devel/gmake, shell/bash, lib/libBlocksRuntime

Snapshot this VM image and play with honggfuzz.

Regarding toolchain with sanitziers, this is an ongoing effort and my focus is the upstream trunk. Please ignore the sanitizers available in the distribution as of now.  This is an ongoing process of streamlining these tools to the Linux level ( https://wip.pkgsrc.org/cgi-bin/gitweb.cgi?p=pkgsrc-wip.git;a=blob;f=compiler-rt-netbsd/TODO.check;h=d5fa535ff188f95a54ddcb46264a5228b18ad4a6;hb=HEAD ). I'm using with @plusun a local pile of patches.

I'm working on micro-UBSan and if I will manage to get some spare time during this month, I will try to port ptrace(2) in honggfuzz to NetBSD. I've managed to run regression tests against uUBSan and catch thousands of reports.. http://netbsd.org/~kamil/mksanitizer-reports/atf-mklibcsanitizer-2018-07-25.txt this means I'm a little bit swamped.

(BTW. mail has been sent)

krytarowski:
@robertswiecki hi! A follow up of this thread.

Is it reasonable/possible to combine `LLVMFuzzerTestOneInput` and charge extra randomness from `HF_ITER` on demand?

Also a version of HF_ITER that has returns the requested number of bytes is wanted.

krytarowski:
We are working on a follow up of the previous work and we would like to charge the fuzzed program on demand for the exact number of bytes from honggfuzz. Would the fuzzer be smart enough to support it?

We would like to avoid collecting a lot of data in a buffer before calling APIs and charge from it manually as it would be too slow and the buffer too large.

krytarowski:
Something like `HF_ITER_INITIAL()` for the initial charge of the buffer and `HF_ITER_CHARGE()` to pull on demand random bytes; both serving the exact number of requested bytes.

krytarowski:
We experiment with something like:

```
static pthread_mutex_t mtx = PTHREAD_MUTEX_INITIALIZER;
static void
HF_MEMGET(void *dst, size_t len)
{
    static uint8_t *buf;
    static size_t buflen;
    size_t diff;

    if (len == 0)
       return;

    pthread_mutex_lock(&mtx);
    do {
        if (buflen == 0) {
            HF_ITER(&buf, &buflen);
            continue;
        }
        diff = MIN(buflen, len);
        memcpy(dst, buf, diff);
        buf += diff;
        buflen -= diff;
        dst += diff;
        len -= diff;
    } while (len > 0);
    pthread_mutex_unlock(&mtx);
}
```

robertswiecki:
The interface ```HF_ITER/LLVMFuzzerTestOneInput``` is generally in charge of delivering bytes/length because it presumably has some strategy about what data should be fuzzed (also structure).

I don't mind adding fetching just random bytes if it serves some purpose, so I guess we could add your implementation of ```HF_MEMGET``` or something similar.

I guess you'd like though to change how honggfuzz statistics behave in such case, i.e. to increase counters only when HF_MEMGET returns?

krytarowski:
I have no opinion on the counters. I just care whether the fuzzer is smart enough to handle a function like `HF_MEMGET()`.

Another nice addition would be to allow to feed an application using `HF_ITER()`/`HF_MEMGET()` with a reproducer. Is there a simple way?

krytarowski:
A nice addition to `HF_MEMGET()` would be to specify with a mask/flags what type of input is allowed. We have got a use-case of allowing e.g. characters valid in file names or in hex ascii numbers. Squashing all the input from a fuzzer is possible, but looks hackish.

A feature request is to pass input as `argv[]` parameters... but it is possibly for another discussion.

robertswiecki:
This all sounds reasonable, I don't feel personally the case for such set of functions too well, but it's just limitation of my imagination, I trust you on this one. Esp. if it's just some addition/modification of libhfuzz, then feel free to submit something.

The latter suggestions (beyond just ```HF_MEMGET()```) would require me wrapping my mind around it, but I guess we can start with something simpler?

PR #214, opened 2018-08-04
Improve the portability of execvp(3) calls

krytarowski:
Do not use hardcoded "cc" or "c++" as argv[0] passed to a compiler.

Some programs including clang on NetBSD try to determine the name of the
executed program by parsing the argv[0] parameter passed to main().

Add a wrapper hf_execvp() that sets argv[0] to the same value as the
first argument to execvp() with an executable filename.

This issue caused an error that Clang/LLVM was calling GCC through
the "cc" filename, and it was causing a fatal build issue.

### DIFF
--- a/hfuzz_cc/hfuzz-cc.c
+++ b/hfuzz_cc/hfuzz-cc.c
@@ -102,6 +102,11 @@ static bool isLDMode(int argc, char** argv) {
     return true;
 }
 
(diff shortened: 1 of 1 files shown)


krytarowski:
I've left the initial hardcoding argv[0] = "cc" / "c++" as is.

PR #218, opened 2018-08-15
Add ptrace(2) backend for NetBSD

krytarowski:
Model this code after existing Linux backend and try to keep differences
to minimum.

Features and design choices, compared to the Linux code:
 - capstone dependency as a disassembler
 - no bfd support
 - -D_KERNTYPES for register_t type
 - no clone() support
 - no disable ASLR switch (personality)
 - no libunwind support, there is no port of HP libunwind version
 - no hardware assisted profiling aid support
 - traditional PT_ATTACH in place of 'seize' operation
 - poll(2) detecting of signals, instead of sigtimedwait(2) which
   didn't work for unknown reasons
 - no cross-ABI support (i386 on amd64 host)
 - currently only amd64 and i386 support

Missing or incomplete features are subject for improvement.

Tested with NetBSD/amd64 8.99.23

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,6 +137,17 @@ else ifeq ($(OS),Darwin)
     endif
     ARCH_SRCS := $(sort $(wildcard mac/*.c))
     # OS Darwin
+else ifeq ($(OS),NetBSD)
+    ARCH := NETBSD
+
+    ARCH_SRCS := $(sort $(wildcard netbsd/*.c))
+    ARCH_CFLAGS := -std=c11 -I/usr/local/include \
+                   -Wextra -Wno-override-init \
+                   -funroll-loops -D_KERNTYPES
+    ARCH_LDFLAGS := -L/usr/local/include \
+                    -pthread -lcapstone -lrt
+
+    # OS NetBSD
 else
     ARCH := POSIX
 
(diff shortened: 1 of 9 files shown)


krytarowski:
CC: @plusun - if possible, please check whether it works for you.

krytarowski:
Some observations:

 - This code shall be relatively simple to port to FreeBSD (but I've never installed that OS and won't do it myself). FreeBSD also has hp libunwind usable on amd64, as opposed to NetBSD today.
 - The generic code seems to be too much clobbered by linux-specific code, such as reporting linux specific fiels or performing operations that might not make sens such as calling performance counter routines and printing as debug messages them on a !inux host.
 - On the other hand Linux/NetBSD/POSIX code could be merged when applicable.

krytarowski:
Regarding sigwaitinfo(2) and explaining the unknown reason why sigtimedwait(2) does not work, I've prepared a test-case and tested on NetBSD and on hosts where I have a shell account (OpenBSD, FreeBSD, Linux, SmartOS, DragonflyBSD) for cross kernel behavior checks.

An undefined behavior with not installing a signal handler is handled differently in the tested OSes.

http://netbsd.org/~kamil/honggfuzz/sigwaitinfo.c

```
// FreeBSD hands
// NetBSD hangs
// OpenBSD no support for sigwaitinfo(2)
// Linux passes over
// SmartOS passes over
// DragonflyBSD hangs
```

I was trying to add a signal handler and block the expected signals, but for some reason I couldn't get it to work, in the context of honggfuzz.

On the other hand I don't see any issues with poll(2) and it just works.

robertswiecki:
Awesome work!

PR #219, opened 2018-08-15
Print proper Thread Id in logs on NetBSD

krytarowski:
All threads on NetBSD have the same process id and distinct thread id.

Add a conditional compilation switch for NetBSD, to implement __hf_pid()
with NetBSD specific call to return the Thread ID, in NetBSD terminolog
LWP ID (ligth-weight process identification).


krytarowski:
` __hf_pid()` probably shall be named `__hf_tid()` or similar as it was unclear what is the real intention.

PR #222, opened 2018-08-18
Handle NetBSD in generic code switches

krytarowski:
Enhance the NetBSD support in generic code using the linux struct,
and where applicable use the netbsd struct.


krytarowski:
@plusun I think that this is the penultimate change for now. Verifying whether it still works is welcome!

It works for me locally.

plusun:
@krytarowski I've checked it and it works! Sorry for the late reply!

krytarowski:
@plusun thank you!

PR #223, opened 2018-08-20
NetBSD related fixups

krytarowski:
Add more mentions of NetBSD in README.md.
Correct dependency file in Makefile linux/unwind.h -> netbsd/unwind.h.
Restore removed include <sys/types.h> in netbsd/arch.c and netbsd/trace.h
and decorate it with clang-format attributes.


README.md
```diff
@@ -7,10 +7,10 @@ A security oriented, feedback-driven, evolutionary, easy-to-use fuzzer with inte
...
   * It's blazingly fast when in the [persistent fuzzing mode](https://github.com/google/honggfuzz/blob/master/docs/PersistentFuzzing.md)). A simple/empty _LLVMFuzzerTestOneInput_ function can be tested with __up to 1mo iterations per second__ on a relatively modern CPU (e.g. i7-6700K)
   * Has a [solid track record](#trophies) of uncovered security bugs: the __only__ (to the date) __vulnerability in OpenSSL with the [critical](https://www.openssl.org/news/secadv/20160926.txt) score mark__ was discovered by honggfuzz. See the [Trophies](#trophies) paragraph for the summary of findings to the date
-  * Uses low-level interfaces to monitor processes (e.g. _ptrace_ under Linux). As opposed to other fuzzers, it __will discover and report hijacked/ignored signals__ (intercepted and potentially hidden by signal handlers)
+  * Uses low-level interfaces to monitor processes (e.g. _ptrace_ under Linux and NetBSD). As opposed to other fuzzers, it __will discover and report hijacked/ignored signals__ (intercepted and potentially hidden by signal handlers)
```
krytarowski: To be precise this is only about 'trap signals' other ones if masked will be still masked. And even trap signals emitted with raise() or similar and masked will remain masked... so it's not that simple.

krytarowski: Probably the best and correct wording is `hijacked/ignored signals from crashes`.

krytarowski: I will prepare a patch for it.

PR #230, opened 2018-10-17
Fix socketfuzzer test code for interactive mode

prashast:
The interactive mode for the socketfuzzer test code was passing illegal argument to connect so I've fixed that and gcc 5.4 does not have the `fsanitize-coverage` option so changed it to clang instead.


prashast:
I signed it!

robertswiecki:
Thx!

PR #244, opened 2019-01-23
libhfcommon: Do not close twice the same file in files_readPidFromFile

francois-berder:
Signed-off-by: Francois Berder <fberder@outlook.fr>


francois-berder:
I signed it!

PR #250, opened 2019-04-08
Add support for external custom feedback data mutator command

arnow117:
I find there is not exist a custom mutator option for feedback data in honggfuzz, so I add an option for using external custom data mutator, which can be used for mutating files by custom logic instead of mutating by internal mutators. And I think this maybe more useful for add custom [structure-aware](https://github.com/google/fuzzer-test-suite/blob/master/tutorial/structure-aware-fuzzing.md) mutator

### DIFF
--- a/cmdline.c
+++ b/cmdline.c
@@ -274,6 +274,7 @@ bool cmdlineParse(int argc, char* argv[], honggfuzz_t* hfuzz) {
                 .fuzzStdin = false,
                 .externalCommand = NULL,
                 .postExternalCommand = NULL,
+                .feedbackMutateCommand = NULL,
                 .persistent = false,
                 .netDriver = false,
                 .asLimit = 0U,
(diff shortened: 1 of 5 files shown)


robertswiecki:
Nice, thanks!

prashast:
I am not sure I understand what additional purpose is this providing over existing `mutate_cmd` option provide which allows specifying an external utility to fuzz an input. Couldn't this `custom logic` be already specified in the external fuzzer?

arnow117:
> I am not sure I understand what additional purpose is this providing over existing `mutate_cmd` option provide which allows specifying an external utility to fuzz an input. Couldn't this `custom logic` be already specified in the external fuzzer?

Well from the honggfuzz [source code](https://github.com/arnow117/honggfuzz/blob/master-add-support/input.c#L362), mutate_cmd can be used in dry-run mode, but not suitable for feedback mode. This extension can provides the ability of mutating data from external utility directly with current feedback-positive input.

arnow117:
> I am not sure I understand what additional purpose is this providing over existing `mutate_cmd` option provide which allows specifying an external utility to fuzz an input. Couldn't this `custom logic` be already specified in the external fuzzer?

Well from the honggfuzz [source code](https://github.com/google/honggfuzz/blob/master/input.c#L360), mutate_cmd can be used in dry-run mode, but not suitable for feedback mode. This extension can provides the ability of mutating data from external utility directly with current feedback-positive input.

prashast:
Aah I hadn't observed that. Thanks for the reply!

PR #269, opened 2019-08-19
qemu_mode: initial support

thebabush:
PoC for qemu_mode.

I don't like AFL's way of building from a specific tarball + applying patches, but QEMU uses submodules so downloading from github as a zip doesn't work.

The idea for now is that I keep a QEMU repo where the `honggfuzz` branch is the current implementation.
AFL's way looks better for more stable stuff.

Let me know what you think.
If you want a more trusted pipeline, I'll do it like AFL does.

So far the coverage implemented is:

- basic block
- cmp

I'm instrumenting instructions in the disassembler. AFAIK that's the most stable and clean option. For x86 I just instrument `cmp`s and on ARM a bunch of instructions. Unfortunately, t's difficult to understand what needs to be instrumented without some empirical data.

It would be nice to have some CI going on since QEMU is huge and complex but for now it seems to work ok.

### DIFF
--- a/qemu_mode/Makefile
+++ b/qemu_mode/Makefile
@@ -1,0 +1,22 @@
+
+.PHONY: clean qemu_bin
+
+TARGETS ?= x86_64-linux-user
+
+qemu_bin: honggfuzz-qemu/config.status
+	@echo "\nRun \"cd honggfuzz-qemu/ && make\"."
+	@echo "Targets: \"honggfuzz-qemu/*-linux-user/qemu-*\".\n"
+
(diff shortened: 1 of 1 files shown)


robertswiecki:
Hi, thanks, I will pull it as is. I think depending on external repo is fine, all in all it's a trust decision.

As for your code, What do you think about linking the final qemu binary with libhfuzz.a ? You're currently duplicating some code from there, adding a custom helper function for PC and CMP. Could we add it natively to libhfuzz, and later link qemu with libhfuzz.a ?

PR #280, opened 2019-11-04
Added option to export the feedback struct to disk

saelo:
Hi, I added a new commandline option to export the feedback_t struct to disk. I'm using that so I can afterwards fairly conveniently visualize the fuzz coverage of a binary. I'm not sure if this is a useful feature in general, and it might also be somewhat problematic as it's now essentially exporting an implementation detail...


robertswiecki:
Thanks Samuel!

PR #287, opened 2019-11-25
netdriver: Make temporary directory overridable

hansmi:
Commit 627189739 changed the name of the path at which a tmpfs is
mounted for the duration of a main program from "/tmp/FUZZ" to
"/tmp/HFND_TMP_DIR". This change adds a weak-linked function allowing
fuzzers to override the temporary directory, e.g.:

```
int HonggfuzzNetDriverTempdir(char *str, size_t size)
{
   return snprintf(str, size, "/tmp/foo.%d", getpid());
}
```

This is particularily useful when multiple users share a system and
shouldn't each other. A future change could make the code use
mkdtemp(3).

The default path remains "/tmp/HFND_TMP_DIR" and the "/tmp/FUZZ"
directory isn't affected by this change.

### DIFF
--- a/libhfnetdriver/netdriver.c
+++ b/libhfnetdriver/netdriver.c
@@ -66,6 +66,15 @@ static void netDriver_startOriginalProgramInThread(void) {
     }
 }
 
+static void netDriver_mountTmpfs(const char *path) {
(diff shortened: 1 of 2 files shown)


robertswiecki:
Thanks!

PR #288, opened 2019-11-25
sanitizers: Disable ODR violation detection

hansmi:
Commit 0b418dd2 has already disabled [ODR](https://en.wikipedia.org/wiki/One_Definition_Rule) detection for honggfuzz invocations without "--sanitizers". This change does the same for when sanitizers are explicitly enabled. ODR violations aren't a concern for fuzzing, but rather for unit or integration tests.

### DIFF
--- a/sanitizers.c
+++ b/sanitizers.c
@@ -68,6 +68,7 @@
 #define kASAN_COMMON_OPTS        \
     "allow_user_segv_handler=1:" \
     "handle_segv=0:"             \
+    "detect_odr_violation=0:"    \
     "allocator_may_return_null=1:" kSAN_COMMON ":exitcode=" HF_XSTR(HF_SAN_EXIT_CODE)
 /* Platform specific flags */
 #if defined(__ANDROID__)


robertswiecki:
Thanks... I'll pull it in, tho the --sanitizers is a legacy flag, it has been used to support the old sanitizer-coverage=bb|edge, but now it has no bigger purpose. I plan to re-use it for some special purposes (intergration with oss-fuzz). Why would you like to use it? Is it because you thought that it somehow enables better support for asan?

hansmi:
> Is it because you thought that it somehow enables better support for asan?

Somewhat, yes. Not necessarily better, but with options more adjusted to fuzzing.

robertswiecki:
Hm.. I'd suggest dropping --sanitizer in your fuzzing code if it works as such, I don't think it'll help to get better results with it. oss-fuzz maintainers wanted for --sanitizers to leave log files around (so asan log can be read from there), so I'll re-work it in some short time span.

hansmi:
Thank you for the hint. There's no need for this patch then and I'll close the PR.

PR #289, opened 2019-11-27
Support AF_UNIX address formatting

hansmi:
The forthcoming introduction of AF_UNIX support in libhfnetdriver
requires the ability to format such addresses via the
"files_sockAddrToStr" function.


robertswiecki:
Thanks!

PR #291, opened 2019-11-28
netdriver: Add support for non-TCP protocols

hansmi:
Many daemons use Unix domain sockets to communicate with their clients.
This change amends libhfnetdriver with support for all
SOCK_STREAM-compatible address families while retaining the previous
behaviour of probing TCP on IPv4 and IPv6.

A newly introduced, weak-linked "HonggfuzzNetDriverServerAddress"
function allows users of the library to provide their own destination
socket address, e.g.:

```
  int HonggfuzzNetDriverServerAddress(struct sockaddr* addr,
    socklen_t* addrlen)
  {
    struct sockaddr_un saddr = {
      .sun_family = AF_UNIX,
    };
    char tempdir[PATH_MAX];

    if (HonggfuzzNetDriverTempdir(tempdir, sizeof(tempdir)) < 0) {
      return -1;
    }

    snprintf(saddr.sun_path, sizeof(saddr.sun_path), "%s/pipe",
      tempdir);

    memcpy(addr, &saddr, sizeof(saddr));
    *addrlen = sizeof(saddr);

    return 0;
  }
```

With the above function libhfnetdriver will probe the Unix socket named
"pipe" in the temporary directory provided by
"HonggfuzzNetDriverTempdir" until connections are accepted and will then
proceed to fuzz as it would a TCP connection.

### DIFF
--- a/libhfnetdriver/netdriver.c
+++ b/libhfnetdriver/netdriver.c
@@ -37,12 +37,13 @@ static struct {
     int argc_server;
     char **argv_server;
     uint16_t tcp_port;
-    sa_family_t sa_family;
+    socklen_t addrlen;
+    struct sockaddr_storage addr;
 } hfnd_globals = {
     .argc_server = 1,
     .argv_server = initial_server_argv,
     .tcp_port = 0,
-    .sa_family = AF_UNSPEC,
+    .addr.ss_family = AF_UNSPEC,
 };
 
 extern int HonggfuzzNetDriver_main(int argc, char **argv);
@@ -145,18 +146,24 @@ static int netDriver_sockConnAddr(const struct sockaddr *addr, socklen_t socklen
     if (setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, &val, (socklen_t)sizeof(val)) == -1) {
         PLOG_W("setsockopt(sock=%d, SOL_SOCKET, SO_REUSEADDR, %d)", sock, val);
     }
+
+    switch (addr->sa_family) {
+        case AF_INET:
+        case AF_INET6:
 #if defined(SOL_TCP) && defined(TCP_NODELAY)
-    val = 1;
-    if (setsockopt(sock, SOL_TCP, TCP_NODELAY, &val, (socklen_t)sizeof(val)) == -1) {
-        PLOG_W("setsockopt(sock=%d, SOL_TCP, TCP_NODELAY, %d)", sock, val);
-    }
+            val = 1;
+            if (setsockopt(sock, SOL_TCP, TCP_NODELAY, &val, (socklen_t)sizeof(val)) == -1) {
+                PLOG_W("setsockopt(sock=%d, SOL_TCP, TCP_NODELAY, %d)", sock, val);
+            }
 #endif /* defined(SOL_TCP) && defined(TCP_NODELAY) */
 #if defined(SOL_TCP) && defined(TCP_QUICKACK)
-    val = 1;
-    if (setsockopt(sock, SOL_TCP, TCP_QUICKACK, &val, (socklen_t)sizeof(val)) == -1) {
-        PLOG_D("setsockopt(sock=%d, SOL_TCP, TCP_QUICKACK, %d)", sock, val);
-    }
+            val = 1;
+            if (setsockopt(sock, SOL_TCP, TCP_QUICKACK, &val, (socklen_t)sizeof(val)) == -1) {
+                PLOG_D("setsockopt(sock=%d, SOL_TCP, TCP_QUICKACK, %d)", sock, val);
+            }
 #endif /* defined(SOL_TCP) && defined(TCP_QUICKACK) */
+            break;
+    }
 
     netDriver_bindToRndLoopback(sock, addr->sa_family);
 
@@ -167,37 +174,6 @@ static int netDriver_sockConnAddr(const struct sockaddr *addr, socklen_t socklen
         return -1;
     }
     return sock;
-}
-
-int netDriver_sockConnLoopback(sa_family_t sa_family, uint16_t portno) {
-    if (portno < 1) {
-        LOG_F("Specified TCP port (%d) cannot be < 1", portno);
-    }
-
-    if (sa_family == AF_INET) {
-        /* IPv4's 127.0.0.1 */
-        const struct sockaddr_in saddr4 = {
-            .sin_family = AF_INET,
-            .sin_port = htons(portno),
-            .sin_addr.s_addr = htonl(INADDR_LOOPBACK),
-        };
-        return netDriver_sockConnAddr((const struct sockaddr *)&saddr4, sizeof(saddr4));
-    }
-
-    if (sa_family == AF_INET6) {
-        /* IPv6's ::1 */
-        const struct sockaddr_in6 saddr6 = {
-            .sin6_family = AF_INET6,
-            .sin6_port = htons(portno),
-            .sin6_flowinfo = 0,
-            .sin6_addr = in6addr_loopback,
-            .sin6_scope_id = 0,
-        };
-        return netDriver_sockConnAddr((const struct sockaddr *)&saddr6, sizeof(saddr6));
-    }
-
-    LOG_E("Unknown SA_FAMILY=%d specified", (int)sa_family);
-    return -1;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


robertswiecki:
Please give me some time to review it. I'm sure the code is quite correct, I just need to think about the whole configuration API. E.g. should we still rely on weak functions, or maybe use envvars to configure all that, or both.

honggfuzz -E HFND_DEST_ADDR=unix://... 

etc.

hansmi:
> Please give me some time to review it.

Sure, take the time you need.

> honggfuzz -E HFND_DEST_ADDR=unix://...

If I may add my CHF 0.05: The weak-linked function working with `struct sockaddr_*` allows fuzzing via all compatible address families without requiring modifications in the fuzzer and there's no need for a parser. Relying on external environment variables makes it impossible for the fuzzer to specify the path (which may not be known until the program is started, see commit bfed5ea505d0).

If the growing number of weak-linked functions is a concern another option could be to switch to a configuration structure. Pseudo code for a fuzzer:

```c
int HonggfuzzNetDriverConfig(struct honggfuzz_netdriver_config *cfg) {
  static char *tmpdir = "/foo/bar";

  cfg->size = sizeof(*cfg);
  cfg->tmpdir = tmpdir;
  cfg->other_option = 1;

  struct sockaddr_un* addr = (struct sockaddr_un*)cfg->target_address;
  addr->sun_family = AF_UNIX;
  addr->sun_path = "…";
}
```

Binary compatibility is a bigger concern with this variant, hence setting the size to allow libhfnetdriver to work with older/smaller versions of the structure.

robertswiecki:
Makes sense, the only comment I have about your code is that there're no length checks  for struct sockaddr* addr. You pass *addrlength, but it's only used as a return value, and not when modifying sockaddr;

Would it be possible to rewrite it, so it maybe does the following.

```C
socklen_t HonggfuzzNetDriverServerAddress(struct sockaddr** addr) {
   static __thread sockaddr_... sa;
  *addr = &sa;
   sa.blabla = blabla;
  return somelength;
}
```

This way we can avoid the input legtht problem, as it's basically the function owning the struct (which is in TLS, which is static, which can be used out of function scope).

robertswiecki:
I'm sorry Michael, could you hold on with your PR for now. I'm rewriting the code, so it's easier to add your AF_UNIX plug. After that it should be a single function change

robertswiecki:
Ok, done in https://github.com/google/honggfuzz/commit/1ea443f8cf676129ddba5327f41a97f565b0c3a9

Would it be possible to implement your AF_UNIX connections on top of that?

If you'd like default checking for TmpfsDir()/pipe, I think you can add this check after checks for TCP4 and TCP6.

hansmi:
> Would it be possible to rewrite it, so it maybe does the following.
> 
> ```c
> socklen_t HonggfuzzNetDriverServerAddress(struct sockaddr** addr) {
> […]
> ```
> 
> This way we can avoid the input legtht problem, as it's basically the function owning the struct (which is in TLS, which is static, which can be used out of function scope).

I didn't want to make the client code the owner of the address storage and by returning the struct length as the return value it becomes impossible to return an error status (e.g. if the fuzzer has code that can fail for some reason).

> Ok, done in [1ea443f](https://github.com/google/honggfuzz/commit/1ea443f8cf676129ddba5327f41a97f565b0c3a9)
> 
> Would it be possible to implement your AF_UNIX connections on top of that?

I don't know without studying your changes in-depth. It's neigh impossible to say [because the commit message doesn't describe the changes at a high level](https://chris.beams.io/posts/git-commit/).

I know you own the code, but I spent considerable time working out an interface that works for fuzzing AF_UNIX and other non-TCP families without breaking compatibility with existing code and actually tested it (e.g. attempting to set `TCP_NODELAY` on AF_UNIX will fail). `netDriver_waitForServerReady` seems excessively repetitive. There was no clear need for a "rewrite".

robertswiecki:
I'm sorry it turned out this way. My fault!

I'll try to improve on the github log messaging, and on discussing the pull request fronts.

To my defense, I felt like the old code structure was tied too much towards TCP, and even support for TCP/IPv6 was somewhat a hack.

I'll take a look at your original PR one more time, and try to implement the functionality you were looking for on my own, to avoid putting the burden of understanding the new code on you, I hope that's ok.

robertswiecki:
FYI: I believe the needed functionality is now implemented by https://github.com/google/honggfuzz/commit/6814cb6a032ab0963bc37493ec4f7f640389daaf

If you're still interested in this code, please feel free to test it.

FWIW: I think that returning socklen_t might be a better option here. First, it's simpler, and simpler is typically better. I don't think we need error handling there. The HonggfuzzNetDriverServerAddress() should either return 0 (function not interested in setting address), or fail hard with panic() (LOG_F here). There's no recovery from error in this path. We shouldn't be trying to connect to default ports if HonggfuzzNetDriverServerAddress() fails, because it means that the code wanted the netdriver to something very specific, and failed to describe it.

hansmi:
@robertswiecki, thanks for the changes. It'll be a few days before I have time again to work on this.

hansmi:
I've only found one minor issue, patch is in #292.

robertswiecki:
Thanks for verifying!

QQ: Would you mind me changing the default path from ${tmpdir}/pipe to ${tmpdir}/sock or ${tmpdir}/socket

It's a bit bikeshedding on my part, but technically it is a socket and not a pipe.

Also.. we're currently supporting SOCK_STREAM only, and I've seen that samba sometimes uses SOCK_DGRAM with AF_UNIX (e.g. the samba-tool uses it). Do you think that supporting it would be useful?

hansmi:
> QQ: Would you mind me changing the default path from ${tmpdir}/pipe to ${tmpdir}/sock or ${tmpdir}/socket

Not at all. The code I'm working on is going to supply its own socket path anyway.

> Also.. we're currently supporting SOCK_STREAM only, and I've seen that samba sometimes uses SOCK_DGRAM with AF_UNIX (e.g. the samba-tool uses it). Do you think that supporting it would be useful?

Yes, that would also allow fuzzing other protocols, e.g. DNS servers via UDP.

robertswiecki:
@hansmi  Would it be posible to take a look at https://github.com/google/honggfuzz/pull/293 ?

It changes the declaration of HonggfuzzNetDriverServerAddress(), but I think the feature is new enough (and you're probably the only user now) to break the compatibility.

This implements the type/protocol params settings, and gets rid of static __threads vars.

hansmi:
@robertswiecki, I have to come back to this reply:

> If the growing number of weak-linked functions is a concern another option could be to switch to a configuration structure. Pseudo code for a fuzzer:
> 
> ```c
> int HonggfuzzNetDriverConfig(struct honggfuzz_netdriver_config *cfg) {
> …
> }
> ```

In my usecase I want to modify the data before sending it to the target server as the requests contain two separate length integers. I'm currently using the following local change with the fuzzing code providing its own `LLVMFuzzerTestOneInput` calling `HonggfuzzNetDriverTestOneInput` after doing its magic:

```
--- a/libhfnetdriver/netdriver.c
+++ b/libhfnetdriver/netdriver.c
@@ -401,7 +401,14 @@ int LLVMFuzzerInitialize(int *argc, char ***argv) {
     return 0;
 }
 
+int HonggfuzzNetDriverTestOneInput(const uint8_t *buf, size_t len);
+
+__attribute__((weak))
 int LLVMFuzzerTestOneInput(const uint8_t *buf, size_t len) {
+       return HonggfuzzNetDriverTestOneInput(buf, len);
+}
+
+int HonggfuzzNetDriverTestOneInput(const uint8_t *buf, size_t len) {
     int sock = netDriver_sockConnAddr((const struct sockaddr *)&hfnd_globals.dest_addr.addr,
         hfnd_globals.dest_addr.slen, hfnd_globals.dest_addr.type, hfnd_globals.dest_addr.protocol);
     if (sock == -1) {
```

Would you be willing to consider a change from the current individual weak-linked functions to a larger configuration structure? Do you have another idea to curb the growth?

robertswiecki:
Sorry for late reply, I missed this.

The config idea sounds intriguing, it can actually also solve the problem of making local changes to the data (custom mutators?), as it can technically contain a function pointer to a mutator function, like the one that is described here: https://github.com/google/fuzzing/blob/master/docs/structure-aware-fuzzing.md

My personal TCP fuzzing cases don't require such heavy configuration, but I'm totally ok with enhancements, if your use-cases call for it.

PR #292, opened 2019-11-29
Use sockaddr length from fuzzer

hansmi:
Commit 6814cb6a0 made major changes and introduced the
"HonggfuzzNetDriverServerAddress" function which can be defined by
a fuzzer. Unfortunately the connection attempt didn't use the
fuzzer-provided length and thus truncated longer socket address
structures such as "sockaddr_un".


robertswiecki:
Thanks for spotting this. Thanks!

PR #293, opened 2019-12-01
netdriver: make it possible to specify both type (e.g. SOCK_DGRAM) and protocol for the destination address.

robertswiecki:
This requires change to the HonggfuzzNetDriverServerAddress()
declaration, as it needs to enable its users to specify those two
additional params.

Additionally, pass struct sockaddr_storage to
HonggfuzzNetDriverServerAddress(), as this will prevent this func for
owning storage for the used sockaddr. It was not a bug, but this
semantics should be more familiar for C coders.

### DIFF
--- a/libhfnetdriver/netdriver.c
+++ b/libhfnetdriver/netdriver.c
@@ -150,7 +150,8 @@ static int netDriver_sockConnAddr(
     }
     if (addr->sa_family == AF_INET || addr->sa_family == AF_INET6) {
         int val = 1;
-        if (setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, &val, (socklen_t)sizeof(val)) == -1) {
+        if (setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, &val, (socklen_t)sizeof(val)) == -1 &&
+            errno == ENOPROTOOPT) {
             PLOG_W("setsockopt(sock=%d, SOL_SOCKET, SO_REUSEADDR, %d)", sock, val);
         }
 #if defined(SOL_TCP) && defined(TCP_NODELAY)
@@ -339,10 +340,11 @@ static bool netDriver_checkIfServerReady(int argc, char **argv) {
         return true;
     }
 #endif /* defined(SOCK_SEQPACKET) */
-    /* Next, try TCP4 and TCP6 connections to the localhost */
+       /* Next, try TCP4 and TCP6 connections to the localhost */
+    const uint16_t tcp_port = netDriver_getTCPPort(argc, argv);
     const struct sockaddr_in addr4 = {
         .sin_family = PF_INET,
-        .sin_port = htons(netDriver_getTCPPort(argc, argv)),
+        .sin_port = htons(tcp_port),
         .sin_addr.s_addr = htonl(INADDR_LOOPBACK),
     };
     if (netDriver_connAndAssign((const struct sockaddr *)&addr4, sizeof(addr4), SOCK_STREAM, 0)) {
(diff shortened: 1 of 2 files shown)


robertswiecki:
@hansmi please feel free to take a look as this changes the API of libhfnetdriver. Otherwise I'll probably push it in a couple of days as is.

libhfnetdriver/netdriver.c
```diff
@@ -134,11 +140,12 @@ static void netDriver_bindToRndLoopback(int sock, sa_family_t sa_family) {
...
+    const struct sockaddr *addr, socklen_t socklen, int type, int protocol) {
+    int sock = socket(addr->sa_family, type, protocol);
     if (sock == -1) {
-        PLOG_W("socket(type=%d for dst_addr='%s', SOCK_STREAM, 0)", addr->sa_family,
-            files_sockAddrToStr(addr, socklen));
+        PLOG_W("socket(family=%d for dst_addr='%s', type=%d, protocol=%d)", addr->sa_family,
+            files_sockAddrToStr(addr, socklen), type, protocol);
         return -1;
     }
     if (addr->sa_family == AF_INET || addr->sa_family == AF_INET6) {
```
hansmi: With the introduction of arbitrary socket types perhaps the options should be set on all sockets and the warning/error only be emitted if it's supported (`ENOPROTOOPT`?) .

libhfnetdriver/netdriver.c
```diff
@@ -283,72 +287,80 @@ static const char *netDriver_getSockPath(int argc HF_ATTR_UNUSED, char **argv HF
...
-        hfnd_globals.saddr = (struct sockaddr *)&addr4;
-        hfnd_globals.slen = sizeof(addr4);
+    if (netDriver_connAndAssign((const struct sockaddr *)&addr4, sizeof(addr4), SOCK_STREAM, 0)) {
         return true;
     }
-    static __thread struct sockaddr_in6 addr6 = {
+    const struct sockaddr_in6 addr6 = {
         .sin6_family = PF_INET6,
-        .sin6_port = 0,
+        .sin6_port = htons(netDriver_getTCPPort(argc, argv)),
```
hansmi: Could reuse `addr4.sin_port` instead of calling the function twice.

libhfnetdriver/netdriver.c
```diff
@@ -283,72 +287,80 @@ static const char *netDriver_getSockPath(int argc HF_ATTR_UNUSED, char **argv HF
...
-        close(fd);
-        hfnd_globals.saddr = (struct sockaddr *)&addr6;
-        hfnd_globals.slen = sizeof(addr6);
+    if (netDriver_connAndAssign((const struct sockaddr *)&addr6, sizeof(addr6), SOCK_STREAM, 0)) {
         return true;
     }
 
     LOG_I("Honggfuzz Net Driver (pid=%d): Waiting for the TCP server process to start "
-          "accepting connections at TCP4/TCP6 port: %hu or at the socket path: '%s'",
+          "accepting connections at TCP4/TCP6 port: %hu or at the socket path: '%.108s'",
```
hansmi: Don't hardcode the `sun_path` length from Linux: other platforms may vary. Workable pattern for sprintf: `sprintf("… '%*s'", strnlen(sun.sun_path, slen - offsetof(struct sockaddr_un, sun.sun_path)), sun.sun_path)` (see `unix(7)`).

libhfnetdriver/netdriver.c
robertswiecki: Good catch. Done.

libhfnetdriver/netdriver.c
robertswiecki: Done by caching const uint16_t tcp_port=

libhfnetdriver/netdriver.c
robertswiecki: Clever trick. Done

PR #295, opened 2020-01-09
Support compilation on macOS 10.15 (Catalina)

### LINKED ISSUE #294: Support of latest MacOS (10.15)
Compilation fails with "unsupported version" error due to [makefile](https://github.com/google/honggfuzz/blob/571a81f80fff6c47d79a4ca13b755e21e8ebe5b8/Makefile#L81) checks.

axtl:
Apple didn't change the CrashReporter for Catalina (verified via Apple Dev download of the zip file containing these) therefore this change should be sufficient to support 10.15.

Fixes #294


robertswiecki:
Thank you!

PR #307, opened 2020-03-11
Makefile: libunwind 1.2.1 (Ubuntu 19.10) wants liblzma

douglasbagnall:
We see

```
ld.lld: error:
/usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libunwind-ptrace.so:
undefined reference to lzma_stream_buffer_decode
```
and so forth.

The BUILD_OSSFUZZ_STATIC branch doesn't need changing because
pkgconfig adds the `-llzma`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ ifeq ($(OS)$(findstring Microsoft,$(KERNEL)),Linux) # matches Linux but excludes
                             -lopcodes -lbfd -liberty -lz \
                             -Wl,-Bdynamic
     else
-            ARCH_LDFLAGS += -lunwind-ptrace -lunwind-generic -lunwind \
+            ARCH_LDFLAGS += -lunwind-ptrace -lunwind-generic -lunwind  -llzma \
                             -lopcodes -lbfd
     endif
     ARCH_LDFLAGS += -lrt -ldl -lm


robertswiecki:
It seems to be a bug in libunwind's package under ubuntu, changing the libs from the current

```shell
-lunwind-ptrace -lunwind-generic -lunwind
```

to pkgconfig

```
`pkg-config --libs libunwind-ptrace libunwind-generic`
```

doesn't help

```shell
$ CFLAGS="-fuse-ld=lld" make clean all
...
ld.lld: error: /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libunwind-ptrace.so: undefined reference to lzma_stream_buffer_decode
ld.lld: error: /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libunwind-ptrace.so: undefined reference to lzma_index_uncompressed_size
ld.lld: error: /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libunwind-ptrace.so: undefined reference to lzma_index_end
ld.lld: error: /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libunwind-ptrace.so: undefined reference to lzma_index_buffer_decode
ld.lld: error: /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libunwind-ptrace.so: undefined reference to lzma_stream_footer_decode
ld.lld: error: /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libunwind-ptrace.so: undefined reference to lzma_index_size
```

The problem is that lubunwind-ptrace.so doesn't require liblzma

```shell
$ ldd /usr/lib/x86_64-linux-gnu/libunwind-ptrace.so.0.0.0 
	linux-vdso.so.1 (0x00007fffc5ba2000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f38c62f8000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f38c6514000)
```

but 

```
$ ldd /usr/lib/x86_64-linux-gnu/libunwind-generic.so
..
	liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5 (0x00007ff0aacfb000)
..
```

does. If lld is picky about order of shared libraries, we can change the order of dynamic libs manually, from

```lunwind-ptrace -lunwind-generic -lunwind```

to

```
-lunwind -lunwind-generic -lunwind-ptrace
```

```shell
$ CFLAGS="-fuse-ld=lld" make clean all
..all good..
```

Technically it's libunwind's job to require lzma if they use it, either through DT_NEEDED in ELF, or via proper pkgconfig file. I guess we can live with the temporary patch (reversing order of unwind libs), but I wouldn't require -lzma as it might not be required by some version of libunwind, so it'd be an unnecessary dependency.

Thanks for bringing it up, please feel free to test at HEAD or with 5def7f03697d264965743f8677d99b916d0648b6

robertswiecki:
Well.. now ld.gold complains about it. Hm.. I guess the current situation is better, as lld is probably used more frequently than ld.gold, but I'll try to fix it for all at some point.

robertswiecki:
Ok.. I merged it, cause it seems to fix it for all 3 linkers, but I guess the problem is really on the libunwind's side.

Thanks for pointing the problem out and for the patch.

douglasbagnall:
thanks for the quick response. (and for Honggfuzz).

PR #315, opened 2020-04-22
Makefile: fix compilation dependency on macOS

PaulGrandperrin:
On a clean tree, when running make, `mac/mach_excServer.c` and
`mac/mach_excUser.c` were not yet generated by the `mig` command.

They were therefore not included in `ARCH_SRCS` and therefore neither
in `SRCS` nor `OBJS` and then were not compiled and not included in the
final binary, which made the compilation fail.

A previous solution to avoid this problem was to run make twice so that
the generated .c files are correctly included in `ARCH_SRCS` at the
second run.

This patch explicitly adds thoses two files as object file targets when
building on macOS and then ajusts compilation targets to make everything build
correctly.

thanks to @shamatar for reporting the issue:
https://github.com/rust-fuzz/honggfuzz-rs/issues/35

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ ifeq ($(OS)$(findstring Microsoft,$(KERNEL)),Linux) # matches Linux but excludes
 else ifeq ($(OS),Darwin)
     ARCH := DARWIN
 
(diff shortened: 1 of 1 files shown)


robertswiecki:
Thanks, this was a problem for some time.

PR #326, opened 2020-05-07
fix -o synonym for --output

jtpereyda:
The `-o` argument was not working:
```
> /path/to/honggfuzz -z -P -i working-corpus/ -o hftmp -- ./target            
/home/jpereyda/code/honggfuzz/honggfuzz: invalid option -- 'o'
Usage: /home/jpereyda/code/honggfuzz/honggfuzz [options] -- path_to_command [args]
Options:
 --help|-h 
        Help plz..
[...]
```

But `--output` did work:
```
> /path/to/honggfuzz -z -P -i working-corpus/ --output hftmp -- ./target
```

After this code change, the following works:
```
> /path/to/honggfuzz -z -P -i working-corpus/ -o hftmp -- ./target
```

Root cause was a missing entry in the `optstring` argument to `getopt_long`

### DIFF
--- a/cmdline.c
+++ b/cmdline.c
@@ -550,7 +550,7 @@ bool cmdlineParse(int argc, char* argv[], honggfuzz_t* hfuzz) {
     int           opt_index = 0;
     for (;;) {
         int c = getopt_long(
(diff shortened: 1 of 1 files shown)


robertswiecki:
Thank you!

PR #327, opened 2020-05-07
make --output dir if it doesn't exist, as already done with -W

jtpereyda:
Running honggfuzz with -o and -W where neither directory exists:
```
> ls | grep hftmp
> /path/honggfuzz -z -P -i working-corpus/ --output hftmpo -W hftmpW -- ./target
[...]
[2020-05-07T09:47:51-0700][F][11897] main():331 Output directory 'hftmpo' is not writeable: No such file or directory
> ls | grep hftmp
hftmpW
```

Only the -W directory is created.

After code change:
```
> ls | grep hftmp
> /path/to/honggfuzz -z -P -i working-corpus/ -o hftmpo -W hftmpW -- ./target
[...fuzzing output...]
> ls | grep hftmp
hftmpo
hftmpW
```

### DIFF
--- a/cmdline.c
+++ b/cmdline.c
@@ -259,6 +259,11 @@ static bool cmdlineVerify(honggfuzz_t* hfuzz) {
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


robertswiecki:
Thanks!

PR #331, opened 2020-05-12
dockerfile: set noninteractive

adrianherrera:
This stops tzdata asking you to enter timezones and things.


robertswiecki:
Thank you!

PR #343, opened 2020-06-11
Improve NetBSD compatibility

krytarowski:
Stop using install -t as it is GNU specific. Replace it with a
portable construct.

Before reading with ptrace(2) from address 0x0, check whether
the VA0 is mappable into the process, as otherwise the ptrace(2)
call will return EINVAL.

Stop intercepting fork(2) events as they are ignored anyway.

Document PTRACE_POSIX_SPAWN events.

Use PT_SET_SIGPASS to disable notifying the parent about
uninteresting signals. The only interesting ones are crash signals,
SIGABRT and SIGSYS. Everything else is directly passed to the
traced process.


robertswiecki:
Thank you!

Issue #346, opened 2020-06-19
performace issue introduced by `wmb()`


liumuqing:
a `mfence` instruction (`wmb`) is added in `__sanitizer_cov_trace_pc_guard`, in commit f23dbfe1899bcc160055a42e747b993135e90acd.

On my environment, If I comment out `wmb()`, I can gain about 5~10% more runs per second (17000/s -> 19000/s)

If I understand correctly, since we are already using the atomic operation in `trace_pc_gurad`, memory barrier may be not necessary?

robertswiecki:
Yup,

We use it quite indiscriminately (just in case). I should take a look at it, and limit the use to those really necessary cases. Some of them are probably needed though.

PR #350, opened 2020-07-02
delete some wmb calls to get more speed

liumuqing:
This PR is for #346 
I delete some wmb calls which may be unnecessary, since there are atomic operations already.

I use libxml2 for test,  and use `F 1` to make a comparable test:
`./honggfuzz -i in -F 1 -- ./examples/libxml2/before_patch` speed: 13000~15000/s
`./honggfuzz -i in -F 1 -- ./examples/libxml2/after_patch` speed: 15000~17000/s


liumuqing:
@googlebot I signed it!

robertswiecki:
Thanks for the time spent on analysis. Awesome work!

Issue #345, opened 2020-06-18
performance issue when trace-cmp is enabled


liumuqing:
`check64_cb` will be called at every `cmp` instruction, to check if the operand exists in fuzzed binary.

In my case it will takes about 20% cpu.

![Screenshot from 2020-06-18 17-15-44](https://user-images.githubusercontent.com/1625020/85013038-84d53f00-b196-11ea-9c4b-76781d7613c5.png)
![Screenshot from 2020-06-18 19-11-26](https://user-images.githubusercontent.com/1625020/85013659-88b59100-b197-11ea-8e92-a0d47b3e3bc7.png)

I think all 64bit values can be cached when `check64_cb` is called. Then following calls would be very fast. (by binary search or hash table lookup). 

check64_cb: https://github.com/google/honggfuzz/blob/4fff3e8/libhfcommon/util.c#L868

robertswiecki:
Yes.. that's somewhat understandable. 

This mode of looking for constants inside binary yields very good results, so we can accept some CPU penalty.

Though, it can be controlled by this call: https://github.com/google/honggfuzz/blob/9e34b1bc435462e5e568b0ec079ea0769dfed617/libhfuzz/instrument.c#L375

I think 10% cpu penalty (or maybe 5%) would be a good tradeoff for the benefits. Could you try modifying (increasing) arguments to instrumentLimitEvery( )   (there are couple of such calls across this library), untill you hit 5 ot 10% overhead? Then we can implement those values (arguments to instrumentLimitEvery( )) based on your research

liumuqing:
@robertswiecki  thank you for your repley, I didn't notice `instrumentLimitEvery` before...
while I still think `looking for constants inside binary` can be done much faster. 
the binary can be scanned only once, to record all distinct 64/32bit values into a hash set or a sorted vector.

Then `check64_cb` can be much faster, it will just take  O(1) or O(lg(n)).

robertswiecki:
I think we can close it with https://github.com/google/honggfuzz/pull/351 now in. Thank you for working on that.

PR #351, opened 2020-07-03
faster util_32bitValInBinary/util_64bitValInBinary

liumuqing:
This PR is related to #345 

It's still WIP....

I use libxml2 for test, and use `-F 1` to make a comparable test (Intel(R) Core(TM) i7-8750H CPU):
./honggfuzz -i in -F 1 -t 10 -- ./examples/libxml2/before_patch speed: about 15000 to 17000/s
./honggfuzz -i in -F 1 -t 10 -- ./examples/libxml2/after_patch speed: about 21000 to 22000/s

and in single thread mod, it has much more improvement
./honggfuzz -i in -F 1 -t 10 -n 1 -- ./examples/libxml2/before_patch speed: about 4500/s
./honggfuzz -i in -F 1 -t 10 -n 1-- ./examples/libxml2/after_patch speed: about 8000 to 8500/s

The basic Idea is to not scan binary each time, but use a binary search way to determine if a value exists in the binary....

Becasue It need a time-consuming initialize stage to get `sorted values of binary`... So it's only suitable for persistent mode...

Do you have any ideas about where should I start to make this idea work with fork-mode？@robertswiecki

### DIFF
--- a/libhfcommon/util.c
+++ b/libhfcommon/util.c
@@ -832,6 +832,29 @@ static int addrStatic_cb(struct dl_phdr_info* info, size_t size HF_ATTR_UNUSED, 
 lhfc_addr_t util_getProgAddr(const void* addr) {
     return (lhfc_addr_t)dl_iterate_phdr(addrStatic_cb, (void*)addr);
 }
+static uint64_t * values64InBinary = NULL;
+static size_t values64InBinary_size = 0;
+static size_t values64InBinary_cap = 0;
+
+static uint32_t * values32InBinary = NULL;
+static size_t values32InBinary_size = 0;
+static size_t values32InBinary_cap = 0;
+
+static int cmp_u64(const void * pa, const void *pb) {
+    uint64_t a = *(uint64_t*) pa;
+    uint64_t b = *(uint64_t*) pb;
+    if (a < b) return -1;
+    if (a > b) return 1;
+    return 0;
+}
+
+static int cmp_u32(const void * pa, const void *pb) {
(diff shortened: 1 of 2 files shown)


liumuqing:
BTW, the original way (scan whole binary) also invalid CPU cache, slowing down the overall performance.
The new method only access a few memory.
That may be the reason that it have more improvement, comparing the previous research mentioned in #345.

liumuqing:
I change some of the code, so the idea for now is:

1. cache all 32/64 bit readonly-values.
2. when util_xxbitValInBinary is called, it first check v is in read-only value set
3. if it's not read-only set, we return to original way to scan the binary, but only within writable segments.

Using libxml2 test case, it's still have a big speed improvement, 
./honggfuzz -i in -F 1 -t 10 -- ./examples/libxml2/after_patch speed: about 21000 to 22000/s
./honggfuzz -i in -F 1 -t 10 -n 1-- ./examples/libxml2/after_patch speed: about 7800 to 8000/s

robertswiecki:
Thank you, awesome work! 

I needed some time, to wrap my head around the code before merging, so sorry about the delay.

liumuqing:
@robetrtswieckiAs I mentioned before.. this pr is still work in progress...

I just wanted to show my idea. So if you think it's on the right direction, I can continue on fork mode.

This pr has problems for fork-mode fuzzing, due to the time-consuming stage to `collect values, sort them`. This stage takes more than one second on libxml2, but only need to run only once. In fork mode

I think the time-consuming stage can be done in fuzzing engine, instead of in instrumented code.

PR #353, opened 2020-07-10
randomize instrumentLimitEvery

liumuqing:
If we run in fork-and-exec mode, `staticCnt` will starts
from zero for each run. So for example `__sanitizer_cov_trace_cmp4` will
only be executed at the 4096th time...

If the target is small to have 4096 cmp during execution,
`__sanitizer_cov_trace_cmp4` trick will never work.
Even if the target is big, `instrumentLimitEvery` is may still biased....

So I just make it random..


robertswiecki:
Cool, thank you!

PR #359, opened 2020-07-30
Remove redundant F_SETSIG call from process fork

stefanbucur:
Setting the signal on a file descriptor to SIGIO should be redundant, according to the POSIX specification. Moreover, this system call currently prevents Honggfuzz from running inside a gVisor sandbox.


robertswiecki:
Thanks!

PR #370, opened 2020-12-01
Fix socket fuzzer

jmccl:
The specific problem I was running into is that the socket fuzzer was running out of memory when *dynfile->data* was being allocated in *input_addDynamicInput* as *run->dynfile->size* was whatever had been on heap. (There's likely other good reasons to initialize the memory though.)

In addition, remove *--sancov* from the documentation as it's been removed from the code.


robertswiecki:
Thank you

PR #372, opened 2020-12-11
Custom mutator feature

cnheitman:
Hi! Here is a patch that may be useful for others.

Honggfuzz offers a way to plug an external mutator, the `--mutate_cmd` command-line option. The main drawback of this approach is its cost. Honggfuzz creates a new process each time it invokes the external mutator.

This patch addresses this issue. It adds a new option: `--custom_mutator <SHARED-OBJECT>`. When this option is present, Honggfuzz will call the mutator passed as parameter. The performance achieved with this new patch is the same as the Honggfuzz's internal mutators since there is no need to create a new process.

The mutator has to export a function called `custom_mutator` with the following prototype:

```c
void custom_mutator(uint8_t *buf, size_t len);
```

NOTE: Tested on Ubuntu.


robertswiecki:
Closing due to ENO-CLA

PR #373, opened 2020-12-11
Log stats to disk

cnheitman:
Hi! Here's a small patch to log some of the stats to disk that may be useful to others.

It adds a new command line option: `--statsfile <FILENAME>`. It logs the following data in CSV format:

```csv
unix_time, last_cov_update, total_exec, exec_per_sec, crashes, unique_crashes, hangs, edge, pc
```

The logging is done every time new coverage is found.


robertswiecki:
Closing due to ENO-CLA

PR #374, opened 2021-01-13
Added support for IP compression when using --linux_perf_ipt_block

mqf20:
# Add support for IP compression when using Intel PT for hardware-based coverage-feedback fuzzing

When using Intel PT for hardware-based coverage-feedback fuzzing (with the `--linux_perf_ipt_block` option), IP compression is ignored (see Intel® 64 and IA-32 Architectures Software Developer’s Manual Vol. 3C Section 35.4.2.2 "Target IP (TIP) Packet"). As a result, there will be many collisions in the coverage bitmap.

This issue is also mentioned in [this README](https://github.com/nyx-fuzz/libxdc_experiments) by the developers of [`libxdc`](https://github.com/nyx-fuzz/libxdc):

> Hongfuzz: Similarly to PTrix, honggfuzz avoids disassembling the stream and instead only considers the contend of the TIP packages. It should be noted that it seems like the authors ignored TIP compression and hence cause a significant amount of collisions in the bitmap.

To support IP compression, the latest IP emitted by TIP, TIP.PGE and TIP.PGD Intel PT packets is tracked using `last_tip_ip`.

### DIFF
--- a/linux/pt.c
+++ b/linux/pt.c
@@ -92,21 +92,21 @@ inline static uint64_t sext(uint64_t val, uint8_t sign) {
     return val & signbit ? val | mask : val & ~mask;
 }
 
-__attribute__((hot)) inline static void perf_ptAnalyzePkt(run_t* run, struct pt_packet* packet) {
-    if (packet->type != ppt_tip) {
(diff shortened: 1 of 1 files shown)


robertswiecki:
Thank you!

PR #380, opened 2021-02-09
allow --save-smaller to overwrite

douglasbagnall:
Previously we would save the first smaller file, but not the second
and subsequent ones.


douglasbagnall:
sorry for the incomplete previous patch.

(To be honest I wasn't expecting it to get merged without discussion and non-linux implementations).

Issue #381, opened 2021-02-10
Android target under host other than Linux fails


gpoulios:
The HOST_OS variable is not utilized when building third party libs through compile-libunwind.sh and compile-libcapstone.sh resulting in build error.

```bash
File: third_party/android/scripts/compile-libunwind.sh

122: export CC="$NDK"/toolchains/llvm/prebuilt/linux-x86_64/bin/"$ANDROID_NDK_COMPILER_PREFIX""$ANDROID_API_V"-clang
123: export CXX="$NDK"/toolchains/llvm/prebuilt/linux-x86_64/bin/"$ANDROID_NDK_COMPILER_PREFIX""$ANDROID_API_V"-clang++
```

should be:
```bash
File: third_party/android/scripts/compile-libunwind.sh

122: export CC="$NDK"/toolchains/llvm/prebuilt/"$HOST_OS"-x86_64/bin/"$ANDROID_NDK_COMPILER_PREFIX""$ANDROID_API_V"-clang
123: export CXX="$NDK"/toolchains/llvm/prebuilt/"$HOST_OS"-x86_64/bin/"$ANDROID_NDK_COMPILER_PREFIX""$ANDROID_API_V"-clang++
```

and:
```bash
File: third_party/android/scripts/compile-capstone.sh

114: export CC="$NDK"/toolchains/llvm/prebuilt/linux-x86_64/bin/"$ANDROID_NDK_COMPILER_PREFIX""$ANDROID_API_V"-clang
115: export CXX="$NDK"/toolchains/llvm/prebuilt/linux-x86_64/bin/"$ANDROID_NDK_COMPILER_PREFIX""$ANDROID_API_V"-clang++
```

should be:
```bash
File: hird_party/android/scripts/compile-capstone.sh

114: export CC="$NDK"/toolchains/llvm/prebuilt/"$HOST_OS"-x86_64/bin/"$ANDROID_NDK_COMPILER_PREFIX""$ANDROID_API_V"-clang
115: export CXX="$NDK"/toolchains/llvm/prebuilt/"$HOST_OS"-x86_64/bin/"$ANDROID_NDK_COMPILER_PREFIX""$ANDROID_API_V"-clang++
```

PR #382, opened 2021-02-10
android target under OSX / android aarch64 libunwind patches

gpoulios:
This PR fixes the following:

- android build of libunwind and libcapstone under OSX would fail due to hardcoded "linux" in toolchain path (#381)
- libunwind with ptrace enabled for android/aarch64 would fail during runtime for the following reasons:
   1. `HAVE_DECL_PTRACE_POKEUSER` was enabled while `HAVE_DECL_PT_GETREGSET` was disabled. This would cause `unw_init_remote` to return `EBADREG`, as a result making every crash appear unique when it clearly wasn't. This config was correct in previous versions of honggfuzz (file: compile-libunwind.sh) but was removed in ac7df4eddcbf10a870e94933e13405544550bac4 
   2. Preprocessor branch on `HAVE_DECL_PT_GETFPREGS` (file src/ptrace/_UPT_access_fpreg.c) did not properly exclude __aarch64__ resulting in the wrong implementation of `_UPT_access_fpreg` being compiled in.

### DIFF
--- a/third_party/android/patches/libunwind.patch
+++ b/third_party/android/patches/libunwind.patch
@@ -188,7 +188,7 @@ index 6379d8e..8222c02 100644
    return 0;
  }
 diff --git a/src/ptrace/_UPT_access_fpreg.c b/src/ptrace/_UPT_access_fpreg.c
-index e90ec47..50a4c25 100644
+index e90ec47d..ae61d093 100644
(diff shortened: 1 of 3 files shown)


robertswiecki:
Thank you!

PR #386, opened 2021-03-13
FreeBSD cpu activity aggregation fix for FreeBSD.

devnexen:
By default, the cpu shows no activity as the proc filesytem is
not mounted by default on this platform, therefore relying on native api
to aggregate the data.


robertswiecki:
Thank you!

PR #387, opened 2021-03-13
Expanding the cpu activity display change to NetBSD which has

devnexen:
a counter per cpu only.


robertswiecki:
Thank you

robertswiecki:
Thanks!

PR #390, opened 2021-04-14
Fix default/debug build with Bazel on OSX

eustas:
For debug / fastbuild (default) Bazel defines `DEBUG` on OSX.

This is somewhat legacy, but Bazel developers consider that "works as intended".

Without this fix `bazelbuild / rules_fuzzing` project does not build on OSX.


robertswiecki:
Thanks!

PR #395, opened 2021-04-20
chore: add a new flag to be able to set the exit code

drahnr:
Especially useful for CI runs to do a limited number of evolutions but exit as soon as the first crash is detected.


drahnr:
@googlebot I signed it!

robertswiecki (commented):
What about simply returning error code if exitUponCrash enabled? I.e. w/o any additional flags.

drahnr:
While that's a reasonable solution as well (granted it's unique for this purpose), the intention was to not change the existing default behaviour.
I am fine with either.

robertswiecki:
I guess it's fine as is. Thank you!

PR #400, opened 2021-05-08
musl build fix proposal

devnexen:


robertswiecki:
Thanks! I don't know if we'll be able to test it for regressions in the future, but I don't mind those changes.

PR #412, opened 2021-08-01
Solaris/Illumos build implementation proposal.

devnexen:


libhfcommon/log.c
```diff
@@ -181,3 +181,18 @@ int logFd(void) {
...
+    va_list ap;
+    va_start(ap, fmt);
+    vdprintf(fd, fmt, ap);
+    va_end(ap);
+}
+
+int vdprintf(int fd, const char *fmt, va_list ap) {
+    FILE *fp;
+    fp = fdopen(fd, "w");
+    return vfprintf(fp, fmt, ap);
```
robertswiecki, line 196: Is fclose needed here?

robertswiecki (changes requested):

libhfcommon/log.c
robertswiecki: Sorry, I mean fclose would close the fd as well, but aren't we allocating struct for FILE* each time we do fdopen?

robertswiecki: May fdopen/fclose on dup'd fd?

devnexen: tried to avoid using the opaque filling approach here but ok may need fclose indeed

devnexen: tried but that faults.

devnexen: will try using the opaque field instead.

robertswiecki:
Thanks!

PR #424, opened 2021-11-01
util_PinThreadToCPUs implementation for solaris based systems.

devnexen:


robertswiecki:
Thank you!

libhfcommon/util.c
```diff
@@ -140,6 +143,33 @@ bool util_PinThreadToCPUs(uint32_t threadno, uint32_t cpucnt) {
...
+    pid_t p;
+    bool ret = true;
+
+    if (pset_create(&set) != 0) {
+        PLOG_W("pset_create failed");
+        return false;
+    }
+
+    for (uint32_t i = 0; i < cpucnt; i++) {
+        if (pset_assign(set, i, NULL) != 0) {
```
robertswiecki, line 157: I think it should be 

```(start_cpu + i) % num_cpus``` instead of just ```i```

Otherwise all threads will be pinned to the same CPU? Also ```% num_cpus``` to make sure that CPU cores actually exist. 

BTW, Thanks for the patches.

PR #425, opened 2021-11-01
solaris cpu pinning fix: adding start_cpu to assign the cpu.

devnexen:


libhfcommon/util.c
```diff
@@ -154,7 +154,7 @@ bool util_PinThreadToCPUs(uint32_t threadno, uint32_t cpucnt) {
     }
 
     for (uint32_t i = 0; i < cpucnt; i++) {
-        if (pset_assign(set, i, NULL) != 0) {
+        if (pset_assign(set, (start_cpu + i), NULL) != 0) {
             PLOG_W("pset_assign(%" PRIu32 "), failed", i);
```
robertswiecki, line 158: I suggest adding ```% num_cpus``` too. Or, maybe it's solaris' default behavior?

devnexen: no you re right :)

robertswiecki:
Nice, thanks!

PR #427, opened 2021-11-05
util_rndInitThread: on BSD use arc4random api instead

devnexen:
which does not fail and file descriptor-less.


robertswiecki:
Thanks

PR #432, opened 2021-12-19
cmdline: fail if command line can't be parsed

evverx:
It should make it easier to integrate honggfuzz into CI.

Trying to integrate honggfuzz into the elfutils testsuite I
wrote a script running fuzz targets with hongfuzz for a couple
of minutes. It worked with honggfuzz built using the master
branch but when I switched to the latest stable version a test
that should have failed passed unexpectedly. The culprit was
--exit_code_upon_crash (which hasn't been released yet). If would
have been much easier to track that down if honggfuzz had complained
that it failed to parse the command line and exited with a non-zero
exit code.

### DIFF
--- a/cmdline.c
+++ b/cmdline.c
@@ -134,7 +134,7 @@ static void cmdlineHelp(const char* pname, struct custom_option* opts) {
 
 static void cmdlineUsage(const char* pname, struct custom_option* opts) {
     cmdlineHelp(pname, opts);
-    exit(0);
+    exit(EXIT_SUCCESS);
 }
 
 bool cmdlineAddEnv(honggfuzz_t* hfuzz, char* env) {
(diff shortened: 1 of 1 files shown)


cmdline.c
```diff
@@ -134,7 +134,7 @@ static void cmdlineHelp(const char* pname, struct custom_option* opts) {
 
 static void cmdlineUsage(const char* pname, struct custom_option* opts) {
     cmdlineHelp(pname, opts);
-    exit(0);
+    exit(EXIT_SUCCESS);
```
robertswiecki, line 137: You'd like EXIT_FAILURE here, no?

robertswiecki: Ah.. ok, I see, thanks.

robertswiecki:
Thanks!

PR #441, opened 2022-01-02
Fix the bug of globalCmpFeedback

Laplace1814:
Fixed a crash when obtaining coverage information due to improper use of globalCmpFeedback, causing honggfuzz to generate invalid test cases

### DIFF
--- a/libhfuzz/instrument.c
+++ b/libhfuzz/instrument.c
@@ -123,6 +123,7 @@ static void initializeCmpFeedback(void) {
             _HF_CMP_BITMAP_FD, sizeof(cmpfeedback_t));
         return;
     }
+    memset(ret, 0, sizeof(cmpfeedback_t));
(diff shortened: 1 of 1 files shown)


libhfuzz/instrument.c
```diff
@@ -123,6 +123,7 @@ static void initializeCmpFeedback(void) {
             _HF_CMP_BITMAP_FD, sizeof(cmpfeedback_t));
         return;
     }
+    memset(ret, 0, sizeof(cmpfeedback_t));
```
robertswiecki, line 126: mmap returns initialized memory (to 0'es). I don't think this is strictly necessary. Does the bug show without this line?

Laplace1814:
hi，
The problem has been found, will resubmit pr later

robertswiecki ***@***.***> 于2022年1月2日周日 18:37写道：

Issue #376, opened 2021-02-01
armv7/ppc64le: error: array subscript 'struct user_regs_struct_64[0]' is partly outside array bounds of 'struct user_regs_struct_32[1]'


rwmjones:
I'm getting a strange compilation error on armv7:

```
linux/trace.c: In function 'arch_traceSaveData':
linux/trace.c:387:46: error: array subscript 'struct user_regs_struct_64[0]' is partly outside array bounds of 'struct user_regs_struct_32[1]' [-Werror=array-bounds]
  387 |         *pc                             = r64->pc;
      |                                           ~~~^~~~
linux/trace.c:319:32: note: while referencing 'regs'
  319 |     struct user_regs_struct_32 regs;
      |                                ^~~~
linux/trace.c:388:46: error: array subscript 'struct user_regs_struct_64[0]' is partly outside array bounds of 'struct user_regs_struct_32[1]' [-Werror=array-bounds]
  388 |         *status_reg                     = r64->pstate;
      |                                           ~~~^~~~~~~~
linux/trace.c:319:32: note: while referencing 'regs'
  319 |     struct user_regs_struct_32 regs;
      |                                ^~~~
linux/trace.c: In function 'arch_traceAnalyzeData':
linux/trace.c:387:46: error: array subscript 'struct user_regs_struct_64[0]' is partly outside array bounds of 'struct user_regs_struct_32[1]' [-Werror=array-bounds]
  387 |         *pc                             = r64->pc;
      |                                           ~~~^~~~
linux/trace.c:319:32: note: while referencing 'regs'
  319 |     struct user_regs_struct_32 regs;
      |                                ^~~~
linux/trace.c:388:46: error: array subscript 'struct user_regs_struct_64[0]' is partly outside array bounds of 'struct user_regs_struct_32[1]' [-Werror=array-bounds]
  388 |         *status_reg                     = r64->pstate;
      |                                           ~~~^~~~~~~~
linux/trace.c:319:32: note: while referencing 'regs'
  319 |     struct user_regs_struct_32 regs;
      |                                ^~~~
```

On ppc64le:
```
linux/trace.c: In function 'arch_traceSaveData':
linux/trace.c:410:46: error: array subscript 'struct user_regs_struct_64[0]' is partly outside array bounds of 'struct pt_regs[1]' [-Werror=array-bounds]
  410 |         *pc                             = r64->nip;
      |                                           ~~~^~~~~
linux/trace.c:321:20: note: while referencing 'regs'
[...]

PR #442, opened 2022-01-03
linux/trace: avoid explicit casts to struct user_regs_struct_*

ldv-alt:
Use union of struct user_regs_struct_32 and struct user_regs_struct_64
to avoid explicit casts to these structures.
Also, do not use struct pt_regs on powerpc and powerpc64.

References: https://github.com/google/honggfuzz/issues/376

This is a PoC, feel free to reuse.

### DIFF
--- a/linux/trace.c
+++ b/linux/trace.c
@@ -125,7 +125,6 @@ struct user_regs_struct_64 {
     uint64_t fs;
     uint64_t gs;
 };
-#define HEADERS_STRUCT struct user_regs_struct_64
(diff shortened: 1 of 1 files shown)


robertswiecki:
Thanks, this is a good idea (the union). I used it in https://github.com/google/honggfuzz/commit/cd77e53e32a2a2cb023dae60c22021d2567a108e in a slightly different implementation. Thanks for the hint.

I tested the chaneg with x86/x86-64/ppc64/arm64/arm32 - but sadly no access to mips for now.

PR #444, opened 2022-01-04
Fix the bug of globalCmpFeedback

Laplace1814:
Fixed the crash problem caused by not initialized when applying for globalCmpFeedback memory. Invalid test cases will be generated due to the crash in the program being tested.


robertswiecki:
Thank you!

PR #455, opened 2022-03-28
Update README.md

mvanotti:
Add a couple of bugs found in [nasm](https://github.com/netwide-assembler/nasm) with honggfuzz


robertswiecki:
Thanks, nice finds!

PR #458, opened 2022-05-01
fuzz binaries inclusion into r/o section.

devnexen:


robertswiecki:
Let's do it, but there's some chance that we'll have to revert it, as I cannot test it under all arch/compiler/linker combinations.

Thanks for the patch

PR #466, opened 2022-06-20
using __builtin_memcpy_inline from LLVM in few places.

devnexen:
Differences with __builtin_memcpy are it does not generate
extra calls and it can work only with compile time sizes.


robertswiecki:
Interesting addition to LLVM, thanks!

PR #467, opened 2022-06-23
Follow-up on #466

devnexen:
Using __builtin_memset_inline from LLVM in few places.
__builtin_memset_inline being more recent, needs to be checked
separetely.


robertswiecki:
Thanks!

PR #470, opened 2022-07-23
util attribute malloc additions.

devnexen:


robertswiecki:
Hi,

I'm ok with this change, though reading through https://stackoverflow.com/questions/18485447/gcc-attribute-malloc it seems that it should be applied to memory in an undefined yet state?

So, probably with _Malloc and with _MMap only, no?

devnexen:
Yes I had heavy hand however I think it still apply to util_Calloc no ?

robertswiecki:
Reading through https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes is confusing - it doesn't use the 'undefined' wording, but focuses on pointer aliasing and non-NULL-ness. I think it's fine for all. I'll leave it open for 24h and probably merge in whatever form you'll leave it, b/c I don't fully understand the description. It seems it can be even applied to fopen().

devnexen:
>  It seems it can be even applied to fopen().

yes it does.

PR #471, opened 2022-08-27
Support macOS Monterey

ChiChou:
* Add more search path for frameworks.
* Also check for macOS 11 and macOS 12
* I didn't test on [macOS Ventura Preview](https://www.apple.com/macos/macos-ventura-preview/) yet so let's see how it goes

I noticed that the previous version checking is based on substring match, which could be incorrect for some cases.


robertswiecki:
Thanks!

PR #475, opened 2022-11-17
Add macOS 13 support

yaroslavyaroslav:
Add new macOS major version condition.

That's actually all of that. Haven't tested it, since I'm on Apple silicon it it failed to build because of arc.


robertswiecki:
Thanks!

PR #479, opened 2022-12-18
libhfuzz add strlcpy wrapper

devnexen:


robertswiecki:
Thanks

libhfuzz/memorycmp.c
```diff
@@ -196,6 +196,16 @@ static inline char* HF_strcpy(char* dest, const char* src, uintptr_t addr) {
     return __builtin_memcpy(dest, src, len + 1);
 }
 
+static inline size_t HF_strlcpy(char* dest, const char* src, size_t clen, uintptr_t addr) {
+    size_t len, rlen = __builtin_strlen(src);
+    len = clen < rlen ? clen : rlen;
+    if (len > 0) {
+        instrumentUpdateCmpMap(addr, util_Log2(len));
+    }
+    (void)__builtin_memcpy(dest, src, len + 1);
```
catenacyber, line 205: @robertswiecki this looks wrong
there is no need for `clen + 1` for strlcpy (it is designed to use the full buffer size accounting for the final NULL already)

This caused false positives on oss-fuzz for Suricata (they do not reproduce with libFuzzer)

catenacyber: Thanks, looks better :-)

Issue #399, opened 2021-04-28
Enhancement: Timeout by cov update time


zakrush:
Hi. 
Can you add flag for setup timeout  by cov update time.

I really need to stop the fuzzing when coverage don't changes for some time.

PR #487, opened 2023-04-12
Support exit_on_time option: issue #399

apach301:
Hi! This PR implements exit-on-time option (alike in afl++) which allows to stop fuzzing session when there is no new coverage discovered for a certain amount of time. This functionality is used for automatic fuzz testing organizing.
There is also an issue #399 requesting this feature.


robertswiecki:
Thank you!

Issue #497, opened 2023-08-11
Unable to compile due to bfd.c


gusti6:
This is in Linux Ubuntu 22.04. I have installed sudo apt-get install binutils-dev libunwind-dev libblocksruntime-dev clang
make

Any recommendations?

Thanks!

Below is the error
```
cc -c -O3 -mtune=native -funroll-loops -std=c11 -I/usr/local/include -D_GNU_SOURCE -Wall -Wextra -Werror -Wno-format-truncation -Wno-override-init -I. -D_FILE_OFFSET_BITS=64 -finline-limit=4000 -D_HF_ARCH_LINUX  -o linux/bfd.o linux/bfd.c

linux/bfd.c:202:21: error: ‘enum disassembler_style’ declared inside parameter list will not be visible outside of this definition or declaration [-Werror]

  202 |     void* buf, enum disassembler_style style HF_ATTR_UNUSED, const char* fmt, ...) {

      |                     ^~~~~~~~~~~~~~~~~~

linux/bfd.c:202:40: error: parameter 2 (‘style’) has incomplete type

  202 |     void* buf, enum disassembler_style style HF_ATTR_UNUSED, const char* fmt, ...) {

linux/bfd.c: In function ‘arch_bfdDisasm’:

linux/bfd.c:249:9: error: cast between incompatible function types from ‘void (*)(struct disassemble_info *, void *, int (*)(void *, const char *, ...))’ to ‘void (*)(void *, void *, void *, void *)’ [-Werror=cast-function-type]

  249 |         (void (*)(void*, void*, void*, void*))init_disassemble_info;

      |         ^

In file included from ./libhfcommon/util.h:28,

                 from ./honggfuzz.h:39,

                 from ./sanitizers.h:26,

                 from ./linux/unwind.h:27,

                 from ./linux/bfd.h:32,

                 from linux/bfd.c:26:

linux/bfd.c: In function ‘arch_bfdFPrintFStyled’:

linux/bfd.c:204:5: error: ‘va_start’ used in function with fixed arguments

  204 |     va_start(args, fmt);

      |     ^~~~~~~~

cc1: all warnings being treated as errors

make: *** [Makefile:279: linux/bfd.o] Error 1


```

gusti6:
This is a repeat from https://github.com/google/honggfuzz/issues/490 so please disregard

PR #498, opened 2023-09-14
pastis patch for honggfuzz

vanhauser-thc:
This is not my patch, it is functionality added by Quarkslab for their fuzzer orchestration.

It adds two optional features:
  1. dynamic input directory (e.g. a libafl/AFL/libfuzzer corpus directory) from which periodically new inputs are read from
  2. a statistic file that prints some data that can be parsed by scripts
 
IMHO these are very useful features that would improve fuzzing with honggfuzz, so I would like to see them added.

The current method without such a patch is to stop honggfuzz, copy the honggfuzz corpus and the other fuzzer corpus directory together, minimize and restart with that as new inputs. which cost a lot of time and reduces effectiveness in a fuzzing campaign and hence why especially this feature is so useful.

If necessary I can add text to the README etc.

### DIFF
--- a/cmdline.c
+++ b/cmdline.c
@@ -527,6 +527,8 @@ bool cmdlineParse(int argc, char* argv[], honggfuzz_t* hfuzz) {
         { { "export_feedback", no_argument, NULL, 0x10E }, "Export the coverage feedback structure as ./hfuzz-feedback" },
         { { "const_feedback", required_argument, NULL, 0x112 }, "Use constant integer/string values from fuzzed programs to mangle input files via a dynamic dictionary (default: true)" },
         { { "pin_thread_cpu", required_argument, NULL, 0x114 }, "Pin a single execution thread to this many consecutive CPUs (default: 0 = no CPU pinning)" },
+        { { "statsfile", required_argument, NULL, 0x115 }, "Stats file" },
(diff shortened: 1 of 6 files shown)


cnheitman:
Hi! I implemented those features for PASTIS ([dynamic-input](https://github.com/cnheitman/honggfuzz/tree/feature/dynamic-input) and [stats file](https://github.com/cnheitman/honggfuzz/tree/feature/dynamic-input)). I did not made a PR because I was having compiling issues on Honggfuzz master (issue #497). 

I could make the PR if these change are going to be merged.

robertswiecki:
Nice!

robertswiecki:
@cnheitman Would you like to submit something on top of this PR, or instead (or modifying) this PR?

cnheitman:
@robertswiecki I would like to keep the authorship of the PR since I wrote those feature for Quarkslab. 

Can we do something about it? Like revert an make the PR from my fork?

robertswiecki:
@cnheitman - my mistake, reverted

cnheitman:
Thanks! I'll make the PR now.

cnheitman:
Done. Thanks @robertswiecki. And thanks @vanhauser-thc for bringing this patch up.

donpiekarz:
@robertswiecki rewriting history is not nice to people :)

robertswiecki:
@donpiekarz Correct, I'll do better in the future!

PR #502, opened 2023-09-25
add --version cmdline parameter

vanhauser-thc:
There is no easy way to identify which honggfuzz version is installed as apparently the version is only displayed once the fuzzing UI is up.
So I added a --version command that exists to display the version.


robertswiecki:
Great, thanks!

PR #504, opened 2023-11-28
Fix log message not updating total counters

AteChroma0j:
When fuzzing without tty (or with —verbose), and an increase in corpus happens totals are always zero. 

For example:
(…) New:1/2/3/4/5/6 Cur:0/0/0/0/0/0

Fixed:
(…) New:1/2/3/4/5/6 Cur:7/8/9/10/11/12


robertswiecki:
Thank you!

PR #507, opened 2023-12-19
util_rndInitThread Linux modernisation proposal.

devnexen:
getrandom syscall exists since enough time now (around 3.17) and we save one file descriptor usage.

### DIFF
--- a/libhfcommon/util.c
+++ b/libhfcommon/util.c
@@ -35,11 +35,12 @@
 #include <pthread.h>
 #if defined(_HF_ARCH_LINUX)
 #include <sched.h>
+#include <sys/syscall.h>
 #endif /* defined(_HF_ARCH_LINUX) */
 #if defined(__FreeBSD__)
 #include <pthread_np.h>
 #include <sys/cpuset.h>
-#endif /* defined(__FreebSD__) */
+#endif /* defined(__FreeBSD__) */
 #if defined(_HF_ARCH_NETBSD)
 #include <sched.h>
 #endif /* defined(_HF_ARCH_NETBSD) */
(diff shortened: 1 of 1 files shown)


libhfcommon/util.c
```diff
@@ -228,14 +229,19 @@ static __thread uint64_t       rndState[2];
...
-    }
-    if (files_readFromFd(fd, (uint8_t*)rndState, sizeof(rndState)) != sizeof(rndState)) {
-        PLOG_F("Couldn't read '%zu' bytes from /dev/urandom", sizeof(rndState));
+    size_t rrnd = 0;
+    size_t trnd = sizeof(rndState);
+
+    while (rrnd < trnd) {
+        // we are taking the harder road to cope with older linux versions without the getrandom's
+        // wrapper
+        ssize_t len = (ssize_t)syscall(SYS_getrandom, (uint8_t*)rndState + rrnd, trnd - rrnd, 0);
```
robertswiecki, line 238: 1. Handle EINTR via TEMP_FAILURE_RETRY()
2. syscall() is a vararg, you need to cast all arguments to (long) becasue badness can happen in some cases.

Use can use files_readFromFd from libhfcommon/files.c as an example of TEMP_FAILURE_RETRY use.

libhfcommon/util.c
```diff
@@ -235,7 +235,8 @@ static void util_rndInitThread(void) {
     while (rrnd < trnd) {
         // we are taking the harder road to cope with older linux versions without the getrandom's
         // wrapper
-        ssize_t len = (ssize_t)syscall(SYS_getrandom, (uint8_t*)rndState + rrnd, trnd - rrnd, 0);
+        ssize_t len = TEMP_FAILURE_RETRY((ssize_t)syscall(
+            SYS_getrandom, (uintptr_t)((uint8_t*)rndState + rrnd), (uintptr_t)(trnd - rrnd), 0));
```
robertswiecki, line 239: 0L :)

robertswiecki: Also, this code is used directly by POSIX code, I think you need to leave the /dev/urandom path, and add getrandom as Linux specific (ifdef HF_ARCH_LINUX)

robertswiecki:
Thank you for this code.

For the posterity, I don't think it's essentially needed, this function is not called frequently. But it has some advantages: no need to open-fd, no need for /dev/urandom to be present (which might not be in some Linux namespaces).

PR #509, opened 2024-01-04
Fix dry run

AteChroma0j:
Currently in dry run phase, honggfuzz partially reads files and mangle them, which is more wet than dry. This PR fixes that behaviour.


robertswiecki:
Thanks

PR #514, opened 2024-11-13
libhfcommon/util: Don't use pthread_once in rnd

mkj:
Since this is thread local storage it can set the value directly rather than needing to go through pthread_once. This gives a small speedup, perhaps 4%.


robertswiecki:
Thanks!

PR #515, opened 2024-11-16
Update README.md

noob6t5:
Website of **QuickFuzz** is down .

There's no reference in website .
So In this PR , Website link is changed with **Github Repo** of QuickFuzz


robertswiecki:
Thank you!

PR #517, opened 2025-02-25
Check coverage for imported seeds

apach301:
Measure usefulness of imported seeds before performing mutations on them. 

By default all imported seeds (from external fuzzer via `--dynamic_dir`) got immediately saved to dynamic corpus (with files dumped on disk in corpus directory) without any coverage checks. Such behavior leads to some problems:

- if external fuzzer generates too much files, it leads to a very large corpus size.
- a lot of imported files with default coverage are spam the dynamic queue, where it may push back more preferable seeds and make useless inputs to be mutated many times;
- there is no feedback on how much imported files turned out to be helpful (AFL++ and libFuzzer could track that).


robertswiecki:
Thanks!

PR #520, opened 2025-03-18
Include corpus queue size in stats file

simonresch:
Adds the size of the dynamic corpus to the stats file. This metric is otherwise only output in the TTY output.


robertswiecki:
Thank you!

Issue #477, opened 2022-11-18
Does not support Apple Silicon


yaroslavyaroslav:
It fails to build because of there's no build case for Apple silicon.

```log
make
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -c -O3 -mtune=native -funroll-loops -std=c11 -I/usr/local/include -D_GNU_SOURCE -Wall -Wextra -Werror -Wno-format-truncation -Wno-override-init -I. -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.0.sdk -x objective-c -pedantic -fblocks -Wno-unused-parameter -Wimplicit -Wunused -Wcomment -Wchar-subscripts -Wuninitialized -Wreturn-type -Wpointer-arith -Wno-gnu-case-range -Wno-gnu-designator -Wno-deprecated-declarations -Wno-unknown-pragmas -Wno-attributes -Wno-embedded-directive -Wno-initializer-overrides -Wno-unknown-warning-option -Wno-gnu-empty-initializer -Wno-format-pedantic -Wno-gnu-statement-expression -mllvm -inline-threshold=2000 -D_HF_ARCH_DARWIN -fblocks -o mac/arch.o mac/arch.c
mac/arch.c:699:5: error: unknown type name 'x86_thread_state_t'; did you mean 'arm_thread_state_t'?
    x86_thread_state_t* platform_in_state = ((x86_thread_state_t*)(void*)in_state);
    ^~~~~~~~~~~~~~~~~~
    arm_thread_state_t
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.0.sdk/usr/include/mach/arm/thread_status.h:101:36: note: 'arm_thread_state_t' declared here
typedef _STRUCT_ARM_THREAD_STATE   arm_thread_state_t;
                                   ^
mac/arch.c:699:66: error: expected expression
    x86_thread_state_t* platform_in_state = ((x86_thread_state_t*)(void*)in_state);
                                                                 ^
mac/arch.c:699:47: error: use of undeclared identifier 'x86_thread_state_t'
    x86_thread_state_t* platform_in_state = ((x86_thread_state_t*)(void*)in_state);
                                              ^
mac/arch.c:699:68: error: expected expression
[...]

PR #521, opened 2025-03-21
Support latest MacOS versions

### LINKED ISSUE #519: Support for MacOS version 14 and above
Currently honggfuzz only supports MacOS version till 13.x . Given that the current major MacOS release version is 15.3.x , there is a need for support for greater MacOS versions.

kyakdan:
Honggfuzz currently only supports MacOS until version 13 and only the X86-64 architecture. The reason is the reliance on `crashwrangle` for crash analysis, which is unavailable for the new ARM-based chips and seems to be discontinued. This PR adds support for all MacOS versions using the POSIX route. More specifically, the PR:
- enables building honggfuzz on MacOS with `OS=POSIX make all`: The Makefile is adjusted to include a preprocessor define for the original build system. For example, this is necessary to enable `hfuzz-cc` to ignore certain linker flags that are not compatible with the MacOS X linker.
- disables ASLR in the child process on MacOS: This is necessary for correct error deduplication through consistent stack hashes.
- undeprecates the `--sanitizers` flag: This flag is then needed for the sanitizer to properly handle signals such as `SIGSEGV` and dump a proper sanitizer log that can be used to compute stack hashes.

This PR fixes #519 and #477

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,7 +209,8 @@ LNETDRIVER_OBJS := $(LNETDRIVER_SRCS:.c=.o)
 LNETDRIVER_ARCH := libhfnetdriver/libhfnetdriver.a
 
 # Respect external user defines
-CFLAGS += $(COMMON_CFLAGS) $(ARCH_CFLAGS) -D_HF_ARCH_${ARCH}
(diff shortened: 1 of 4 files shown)


robertswiecki:
Thank you!

PR #524, opened 2025-04-07
Ensure libhfuzz is linked to executables

kyakdan:
We encountered several cases where the fuzz tests were not compiled properly for the persistent mode. The reason for that is that certain symbols were not linked into the final executable. This includes the `main` function and the `"\x01_LIBHFUZZ_PERSISTENT_BINARY_SIGNATURE_\x02\xFF"` signature checked by honggfuzz to determine whether the fuzz test was compiled for the persistent mode. 

This PR fixes this by forcing the linking of `libhfuzz` to all executables. This is inspired by how libFuzzer linked the fuzzer runtime when using `-fsanitize=fuzzer`. This works when linking libhfuzz to normal executables (not fuzz tests) since `main` is declared as a weak function.

### DIFF
--- a/hfuzz_cc/hfuzz-cc.c
+++ b/hfuzz_cc/hfuzz-cc.c
@@ -127,6 +127,35 @@ static bool isLDMode(int argc, char** argv) {
     return true;
 }
 
+static bool isExecutableBuild(int argc, char** argv) {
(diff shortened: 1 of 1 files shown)


robertswiecki:
Thanks for spotting this!

PR #527, opened 2025-05-07
Fix linking of fuzz test on mac

simonresch:
`-all_load` forces all libraries on the linker command line to included which causes problems when linking libhfnetdriver which also contains a `LLVMFuzzerTestOneInput` symbol. With `-force_load` only the libhfuzz library is linked retaining all symbols.


robertswiecki:
Thank you!

PR #536, opened 2026-03-14
Makefile: allow using BUILD_DIR for outputs

PaulGrandperrin:
Hi!

We recently got an issue (https://github.com/rust-fuzz/honggfuzz-rs/pull/116) with `honggfuzz-rs` when building `honggfuzz` from a read-only directory. This happens for example when using Nix as a build system.

Rust's cargo usually writes all its artifacts in a `target` directory but until now my `honggfuzz-rs` project didn't respect that and built `honggfuzz`'s objects files directly in the source directory.

This violates cargo's mechanism and breaks Nix builds.

One easy solution is to copy the whole `honggfuzz` project in the `target` directory each time and build from here. I think this is not great as `honggfuzz` is a hundred or so MBs and cargo will duplicate it as many times as there will be build configurations.

We could avoid copying the `examples` dir or using symlinks/hardlinks but I think the best solution would be have the `Makefile`  be able to use a separate output directory, similarly to cmake.

To illustrate the idea I send you this PR but don't hesitate to reject it, I didn't even write it as I only know the basics of Makefiles. This code was written by Claude Opus 4.6 and I don't fully understand it. It seems to work fine though.

## Summary of Changes (written by Claude)

The Makefile now supports an optional `BUILD_DIR` variable for out-of-tree builds. When not set, it builds in-tree exactly as before.

### Key mechanisms:

1. **`SRCDIR`** — automatically derived from the Makefile's own location using `$(abspath $(lastword $(MAKEFILE_LIST)))`, so it always points to the source tree regardless of where `make` is invoked from.

2. **`BUILD_DIR`** — user-supplied, optional. When set, all build artifacts (`.o`, `.a`, `.so`, binaries) go there instead of the source tree.

3. **`_OBJDIR`** — internal variable that equals `BUILD_DIR` when set, or `SRCDIR` when not (backward compat).

4. **`VPATH := $(SRCDIR)`** — lets Make find `.c` source files in the source tree when building from a different directory.

5. **`-I$(SRCDIR)`** — replaces the old `-I.` so header includes always resolve against the source tree.

6. **`.SECONDEXPANSION:`** with `| $$(dir $$@)` order-only prerequisites — ensures build subdirectories (`linux/`, `libhfcommon/`, etc.) are created in `BUILD_DIR` before compilation.

7. **All output paths** (`OBJS`, `BIN`, `HFUZZ_CC_BIN`, `LHFUZZ_ARCH`, `LCOMMON_ARCH`, `LNETDRIVER_ARCH`, etc.) are prefixed with `$(_OBJDIR)/`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,15 +20,38 @@
 #   NOTE: xcrun is within xcode...xcode is required on OSX.
 #
 
+# --- Out-of-tree build support ---
+# Set BUILD_DIR to build in a separate directory. When unset, builds in-tree
+# as before. Example: make -f /path/to/honggfuzz/Makefile BUILD_DIR=/tmp/build
+#
+# SRCDIR: absolute path to the source tree (where this Makefile lives)
+SRCDIR := $(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
+
+# BUILD_DIR: where build artifacts go. Defaults to SRCDIR (in-tree build).
+BUILD_DIR ?=
+
+# When BUILD_DIR is set, use it; otherwise everything is relative to SRCDIR
+# (preserving original behaviour).
+ifdef BUILD_DIR
+  # Resolve to absolute path and strip trailing slash
+  override BUILD_DIR := $(patsubst %/,%,$(abspath $(BUILD_DIR)))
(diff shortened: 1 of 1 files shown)


robertswiecki:
Thank you Paul!

PR #538, opened 2026-06-13
Makefile: fix linking when using BUILD_DIR

PaulGrandperrin:
I'm integrating the new feature of my last PR into `honggfuzz-rs` and it seems I missed a small patch on my last PR.

Without this patch, the linking fails when using BUILD_DIR.


robertswiecki:
Thank you!

